PR #12, opened 2024-08-03
feat(wifi): Add Ethernet connection icon

thomasroodnl:
Adds the ability to show an ethernet connection icon when connected to ethernet, instead of showing 'no wifi' icon. Since `netsh` seems to give problems for the lan interfaces, I use the winsdk to identify whether the current connection is WiFi or not. If it is WiFi or disconnected, it uses the `netsh` code, otherwise sets an Ethernet icon.

I did switch to Segoe Fluent Icons because I could not get the NerdFont icons to look nice for the ethernet one. If we want the icons to stay the same, we'd need to look a bit deeper into it.

### DIFF
--- a/src/config.yaml
+++ b/src/config.yaml
@@ -174,12 +174,6 @@ widgets:
       label: "<span>{wifi_icon}</span>"
       label_alt: "{wifi_name} {wifi_strength}%"
       update_interval: 5000
-      wifi_icons:
-        - "\udb82\udd2e" # 0% strength (no wifi)
-        - "\udb82\udd1f" # 1-25% strength
-        - "\udb82\udd22" # 26-50% strength
-        - "\udb82\udd25" # 51-75% strength
-        - "\udb82\udd28" # 76-100% strength. Alternate theming: \uf1eb
       callbacks:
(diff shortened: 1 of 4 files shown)


amnweb:
There is a problem with `Segoe Fluent Icons`. It looks really bad.

Segoe 
Screenshot 2024-08-03 135424

Nerd
Screenshot 2024-08-03 135451


Also, maybe this one `\ueba9` (nerd fonts) can be used for ethernet?

Or simply we can switch class names for icons under `update_label` so we can have separate icons; I mean, we can use different font-family.

thomasroodnl:
I agree with you, but I am not sure how to style it. The ones I tried with nerd fonts (e.g. \udb80\ude00) looked bad for some reason and I could not get it to behave properly. I like the nerd font WiFi one better as well, so if you can find a nerd font icon for ethernet that looks nice I am all for keeping it that way :).

thomasroodnl:
How about this one? 
image

It's not really ethernet but close enough for me. It is \udb84\udf94. Border radius only kicks in at `font-size: 20px` though.

PR #13, opened 2024-08-06
feat(media): Main/subtext label, thumbnail update fix, lightness thumbnail

thomasroodnl:
Big update with a lot of neat features:
* Add two vertical text fields, allowing title and artist to be listed above each other, making the widget smaller with more info! The labels can be configured through the config.
* Fixed a bug where the thumbnail would sometimes not update correctly on song change, as the title and artist fields would change before the thumbnail, causing a thumbnail update to the same old thumbnail. Now, once artist and title change, it will update the thumbnail until it is different from the previous one, and then stops updating and waits for title/artist change.
* Thumbnail alpha is now calculated based on the lightness of the image, which should help equalize the 'darkness' of the images a bit. This means that bright thumbnails are darkened more than light thumbnails, hopefully helping with readability. Two parameters, `thumbnail_alpha_range` and `thumbnail_alpha_multiplier`, control this calculation.

Also removes the `start_timer()` overload, like we discussed :)

### DIFF
--- a/src/config.yaml
+++ b/src/config.yaml
@@ -68,8 +68,8 @@ widgets:
       controls_only: false
       controls_left: false
       hide_empty: true
-      thumbnail_alpha_range: 0.75 # Range in which image lightness controls alpha. At 0, alpha fully controlled by multiplier
-      thumbnail_alpha_multiplier: 0.8 # Scales the thumbnail alpha as a whole (after lightness calculation)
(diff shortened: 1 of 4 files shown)


thomasroodnl:
I am closing this PR because I am doing substantial work on the MediaWidget back-end, which has priority. I will open a new PR with this functionality on the updated version once it is done.

Issue #14, opened 2024-08-08
Add placeholder content while fetching info


coffandro:
**What is the request?.**
I'm not on the fastest PC so sometimes it can take some time (like 10 sec) before yasb begins displaying stuff, I would love if it showed a blocky version of the bar or just some placeholder content (like a "getting data" label) before fetching content to display that it's at least kinda working

**Why is it useful?**
It would allow for showing the user that yasb is in fact working in the background.

**How could it be implemented?**
I would think maybe starting by creating the sys tray icon, then the gui and then *only* when that's present begin getting data in another thread

**Additional context**
N/A

thomasroodnl:
I think this is a good idea, I have had a similar experience with the older yasb fork when I had some widgets that were slow (e.g. because they were doing network calls). Once media is stable I could have a go at this.

PR #15, opened 2024-08-09
feat(launch): Loading indication on startup

thomasroodnl:
Previously, the bar would force update each widget before showing the UI. On slower systems (or heavier widgets), this could cause the bar to be invisible for some time, making it unclear for the user that yasb is running. Deferring the initial update to an instantaneous background call should help make the bar visible sooner.

Default text for many widgets is set to 'Loading', further indicating to the user that the widgets are loading.


thomasroodnl:
Linked to issue: https://github.com/amnweb/yasb/issues/14#issue-2455258237

PR #20, opened 2024-08-16
update(readme): added extra options to how to run section.

ashwinjadhav818:
Added a section to use yasb with the installer given in the releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,9 @@
 - Look for the latest release version, which will typically be listed at the top.
 - Under the "Assets" section of the release, you’ll find various files. Click on the installer file to download it.
 
+> [!NOTE]  
+> The installer is currently is unstable and crashes when config.yaml is changed. 
+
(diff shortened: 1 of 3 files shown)


amnweb:
Maybe add some **note** about the installer that is still unstable and crashing can be expected when the config is changed?
Also, I'm tired and can't thinking :( but please let me know why we need to import function_map inside base?

ashwinjadhav818:
Added a note about the installer being unstable.

The function_map is needed inside base so that in-build widgets like volume, wifi, battery can also use the functions on their callbacks not only custom widgets.

amnweb:
So if you call `quick_settings` from callbacks `left_click`, it should be triggered from all widgets. Did you try this?
Something like this ?

```
callbacks:
  on_left: "quick_settings"
```

ashwinjadhav818:
yeahh but it should be `exec quick_settings`

PR #34, opened 2024-09-13
Media widget improvements

aaltofar:
### Apple Music artist fix:
Apple Music delivers media_info in a weird way, where artist is empty, and instead is baked into album_artist along with the song title. This is a fix for this, and it makes the widget work for us Apple Music users.

### Hide media controls:
I just want the media info, so I made it possible to hide them with a hide_controls bool


amnweb:
I don't have Apple Music to test but posted this PR in the Discord group, anyway, since `hide_controls` is not something which is important, maybe it is a good idea to set `required: False` for `hide_controls` in the validation file; in that case, we won't break users' config when they update YASB.

aaltofar:
> I don't have Apple Music to test but posted this PR in the Discord group, anyway, since `hide_controls` is not something which is important, maybe it is a good idea to set `required: False` for `hide_controls` in the validation file; in that case, we won't break users' config when they update YASB.

Sweet, I set the required: false flag now

amnweb:
Thank you.
There is some problem when I set `hide_controls: true`

```
  File "C:\Users\amn\Desktop\l\yasb\src\core\widgets\yasb\media.py", line 150, in _on_playback_info_changed
    self._play_label.setText(self._media_button_icons['pause' if playback_info.playback_status == 4 else 'play'])
    ^^^^^^^^^^^^^^^^
AttributeError: 'MediaWidget' object has no attribute '_play_label'
```

aaltofar:
> Thank you. There is some problem when I set `hide_controls: true`
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\amn\Desktop\l\yasb\src\core\widgets\yasb\media.py", line 150, in _on_playback_info_changed
>     self._play_label.setText(self._media_button_icons['pause' if playback_info.playback_status == 4 else 'play'])
>     ^^^^^^^^^^^^^^^^
> AttributeError: 'MediaWidget' object has no attribute '_play_label'
> ```

Yes, my bad. I forgot to add the checks to changes before commiting

amnweb:
I will test later but what I can see this line can be problem https://github.com/amnweb/yasb/blob/f874999e397630526d4a2138b33c268c01bdde81/src/core/widgets/yasb/media.py#L135 so if you have `_hide_controls` in config active_label won't hide ?

amnweb:
Yeah, tested, and that's what I said in the comment above. If you have `hide_controls: True` in this case, the media widget won't hide.

PR #41, opened 2024-10-02
Added screens option for widgets

ValdamireGido:
Added possibility to select 'all' or 'primary' only screens for a widget to restrict some to primary display only.

ex:
```
  media:
    type: "yasb.media.MediaWidget"
    screens: "primary"
    options:
      ...
```


amnweb:
Just a question, why do we need this if we can set up widgets per monitor? Each monitor can have a different widget and bar.

ValdamireGido:
Well, apparently, I didn't do my investigations properly before making these changes. Just found this in the docs. 

Closing the PR. No need for these changes with the config options you mentioned.

PR #59, opened 2024-10-22
feat(widget): Libre Hardware Monitor Widget

Video-Nomad:
Adds the [Libre Hardware Monitor](https://github.com/LibreHardwareMonitor/LibreHardwareMonitor) (LHM) widget that connects to the local LHM web server and requests the selected sensor data using `SensorId` string.

SensorId can be found in http://localhost:8085/data.json when the LHM server is running.

Can display current value, min, max, unit and histogram.
Value unit is also received from the LHM so `{info[value]} {info[unit]}` will display 52 °C.

Only the specific sensor data is requested.
The post request is done asynchronously using QNetworkAccessManager to avoid locking the UI.

`update_interval` is limited to the LHM update interval.

Min and max values are not possible to get from the LHM, so they are calculated from the history. History size is also adjustable in the config. History is not persistent, so when YASB is reloaded, it is reset.

By default, histogram will use historical min/max for its range, but this can be changed in the config to lock the min and/or max value.

Most common errors are handled and will be displayed in the widget instead of the value - no connection, auth required, invalid sensor id. Errors will not lock the UI.

Auth is supported.

Tested on both the release and the latest source compiled LHM versions.

Known issue:
It seems that sometimes SensorId is not unique. I only encountered it with Battery Level and Degradation Level of the battery. They are both "/battery/level/0" and the widget just displays the degradation level no matter what. Not sure if that's my laptop-specific issue, or just LHM bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ for more themes visit [yasb-themes](https://github.com/amnweb/yasb-themes)
 - **[Github](https://github.com/amnweb/yasb/wiki/(Widget)-Github)**: Shows notifications from GitHub.
 - **[Disk](https://github.com/amnweb/yasb/wiki/(Widget)-Disk)**: Displays disk usage information.
 - **[Language](https://github.com/amnweb/yasb/wiki/(Widget)-Language)**: Shows the current input language.
+- **[Libre Hardware Monitor](https://github.com/amnweb/yasb/wiki/(Widget)-Libre-HW-Monitor)**: Connects to Libre Hardware Monitor to get sensor data.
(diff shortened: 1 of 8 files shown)


amnweb:
Thanks for PR.
For me it looks fine, except for one thing, when there's no connection we need to prevent errors in `_toggle_label` and `_update_label`, maybe simple put same data for label_alt when there is connection error ?


https://github.com/user-attachments/assets/3dda511f-985e-4d7d-9733-4f0357a7dc98

amnweb:
@Video-Nomad thanks.

PR #81, opened 2024-12-10
Add weather location env option

notTamion:
i would like to publish my dotfiles however i can't do that without leaking my location or at least general location. This just allows you to set the weather location via an environment variable


amnweb:
Already added 
`self._api_key = api_key if api_key != 'env' else os.getenv('YASB_WEATHER_API_KEY')`

PR #91, opened 2024-12-27
Add support for using keyboard layout in Language widget

### LINKED ISSUE #90: Keyboard layout in the Language widget
**What is the request?.**
I want to be able to access the keyboard layout from the lang array in the LanguageWidget. 

**Why is it useful?**
Some people use different layouts with the same language, like this:
image

In those cases the current options returned on the lang array, are not enough to see what the current layout is. For example, in the image shown, the language component would only show e.g. `en-GB` in both of those cases, even though en-SE and en-US would give better info that what kind of layout the user currently is writing in.

**How could it be implemented?**
The current objects stored in the lang array contains
```py
{
  'language_code': language_code,
  'country_code': country_code,
  'full_name': full_name,
  'native_country_name': native_country_name.value,
  'native_lang_name': native_lang_name.value
}
```
and that could be extended with `keyboard_layout` which would return the current layout like se-SV or just SV, en-US, US or some other identifier.

Osyx:
Implements a `layout_name` variable accessible via `{lang[layout_name]}` in the Language widget.

Closes #90. 

This config:
```yml
widgets:
  language:
    type: 'yasb.language.LanguageWidget'
    options:
      label: "<span>\uf11c</span>{lang[language_code]}-{lang[country_code]}  {lang[layout_name]}"
```
will give this:
image

### DIFF
--- a/docs/widgets/(Widget)-Language.md
+++ b/docs/widgets/(Widget)-Language.md
@@ -22,7 +22,7 @@ language:
 ```
 
 ## Description of Options
-- **label:** The format string for the label. You can use placeholders like `{lang[language_code]}`, `{lang[country_code]}`, `{lang[full_name]}`, `{lang[native_country_name]}`, `{lang[native_lang_name]}`, `{lang[layout_name]}`.
+- **label:** The format string for the label. You can use placeholders like `{lang[language_code]}`, `{lang[country_code]}`, `{lang[full_name]}`, `{lang[native_country_name]}`, `{lang[native_lang_name]}`, `{lang[layout_name]}`, `{lang[full_layout_name]}`.
 - **label_alt:** The alternative format string for the label. Useful for displaying the full language name.
 - **update_interval:** The interval in seconds to update the language information. Must be between 1 and 3600.
 - **callbacks:** A dictionary specifying the callbacks for mouse events. The keys are `on_left`, `on_middle`, and `on_right`, and the values are the names of the callback functions.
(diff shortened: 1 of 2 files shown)


Osyx:
It seems to work somewhat, but when I'm running it locally it only updates the layout on start, and then never updates it again, even though the rest of the returned object is updated.

Open to suggestions on how to fix! Haven't really written python in ages, and never used Win APIs really, so I'm very rusty.

amnweb:
> Open to suggestions on how to fix! Haven't really written python in ages, and never used Win APIs really, so I'm very rusty.

Welcome to the Windows API jungle xD

amnweb:
I have tried to play with that but I have no idea why `layout_name` is not updating, maybe it is that API limitation or something else not sure :(

Osyx:
@amnweb Hmm that's sad, idk how we should go forward then, I didn't manage to find another way either :/ 

I saw in the docs that they recommended using [Windows.Globalization.Language.CurrentInputMethodLanguageTag](https://learn.microsoft.com/en-us/uwp/api/windows.globalization.language.currentinputmethodlanguagetag) instead. Tried implementing that, but it yielded the same result as `GetKeyboardLayout(thread_id)`, though it also only updated on startup. Also saw this comment:
> This property returns a value for the currently active input method on the thread with keyboard input focus. Calling this on a background thread may not return the expected results.

which is basically the only way to use CurrentInputMethodLanguageTag it seems, so that seems like a dead end.

We could rename it to `initial_layout_name` but might not be worth due to people expecting more of it. 
I think my use case is quite uncommon as well. So maybe we should just close this PR, idk.

Osyx:
Showcase:
[test-yasb-layout.webm](https://github.com/user-attachments/assets/3182185f-d9c4-4f2c-8b20-04882b504a24)

PR #147, opened 2025-03-05
Added overcast, mist, and fog conditions to "cloudy"

KageKaz:
Right now "overcast" is not categorized into a condition. I was wondering if it (along with mist and fog) could be put into cloudy. Or maybe we can make a new category for it?


amnweb:
I'm fine to stay under `cloudy`. Is there anything else we're missing there?

amnweb:
Fixed in https://github.com/amnweb/yasb/pull/277

PR #149, opened 2025-03-08
Fixed taskbar not showing all apps icons

arthurferrai:
This pull request changes taskbar icon rendering so it will look to all open programs and render all icons except the windows that won't show neither on alt-tab nor task manager (it does so by checking for `WS_EX_NOREDIRECTIONBITMAP` style, which is not included on `win32con`).

It also fixes some typing and changes some checks.

With this pull request, all taskbar icons will show even if you start yasb after a bunch of programs are opened before.

### DIFF
--- a/src/core/widgets/yasb/taskbar.py
+++ b/src/core/widgets/yasb/taskbar.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Optional
 from settings import DEBUG
 from core.widgets.base import BaseWidget
 from core.utils.win32.windows import WinEvent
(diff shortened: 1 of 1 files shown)


amnweb:
@arthurferrai thanks for this PR. This is what we need. I have just made some changes in the taskbar widget to get better icon quality. Can you check your code once again, please?

Osyx:
Closes #150

arthurferrai:
> @arthurferrai thanks for this PR. This is what we need. I have just made some changes in the taskbar widget to get better icon quality. Can you check your code once again, please?

I've merged main into this branch and tested here, and it works as expected.

amnweb:
Thanks, it works fine for me as well.

PR #151, opened 2025-03-09
Added extra class to taskbar foreground app

arthurferrai:
With this PR it is possible to add extra style to foreground application using class `.app-icon.foreground`.


amnweb:
Thank you for the PR, it’s another great one.

Issue #139, opened 2025-02-16
Switch between multiple media sources in Media widget.


b1scu1t:
### What is the request?

Currently, the media widget would follow Windows' 'focused' media application, which can sometimes change at random or be hijacked applications like Chrome. 

For example, Chrome has a paused video but Foobar2000 is currently playing music. Foobar2000 is the last application to be paused or played, but if the user selects "Play" via the Media Key or the Media Widget in Yasb, Chrome is always focused and targeted. I doubt Yasb can change what Windows prefers, but an option to switch the media source in the Media widget can significantly improve it. In the picture below, Windows' volume popup allows the user to switch to another media source.

![Image](https://github.com/user-attachments/assets/70e6cf79-d9f2-450f-84f2-9cfcb69dbe2a)

Another third-party app called [Media Flyout](https://github.com/krlvm/MediaFlyout) would show multiple media sources. It doesn't "select" a source, but the "selector" could look like it.
![Image](https://github.com/user-attachments/assets/20453609-50f4-4103-b987-a7017cc5cd52)

### Why is it useful?

Allows the user to select the media source for the Media widget to display and control.

Dusk-afk:
@amnweb I'd love to work on this feature. I am thinking of using scroll wheel event to do the switching. What do you say?

Dusk-afk:
> > [@amnweb](https://github.com/amnweb) I'd love to work on this feature. I am thinking of using scroll wheel event to do the switching. What do you say?
> 
> Do you mean when the mouse cursor is over the widget and using the scroll wheel to switch source?

Yes

amnweb:
@b1scu1t can you test this PR https://github.com/amnweb/yasb/pull/162

@Dusk-afk I have tested and encountered a bunch of errors. I cleared the terminal and ran the process again, but I couldn't reproduce the same issue. I will continue testing to determine where these errors are coming from.

PR #162, opened 2025-03-20
feat(media): add session switching functionality

Dusk-afk:
This pull request adds session switching functionality to the media widget, addressing #139.
Using scroll wheel on the media widget, sessions can be switched.
- Scroll down = Next session
- Scroll up = Previous session

#### Changes:
* Added `_current_session_only` decorator to ensure session update functions are called only if the session is the same as the current session.
* Added session switching logic with `switch_session` method and updated `play_pause`, `prev`, and `next` methods to use the current session.
* Added logic to hide the thumbnail if there is no media in the session in `_on_media_properties_changed`.
* Implemented `wheelEvent` to switch media sessions using the mouse wheel.

#### Demo:
Following is a demo of the media widget handling 3 sessions:
1. Apple Music
2. Spotify
3. YT Music (Browser)

https://github.com/user-attachments/assets/7e8a80ab-4bb7-4ff6-9539-10d1e6f0e73d

### DIFF
--- a/src/core/utils/win32/media.py
+++ b/src/core/utils/win32/media.py
@@ -1,4 +1,3 @@
-import ctypes
 import logging
 from settings import DEBUG 
 from typing import Any, Callable
@@ -11,7 +10,6 @@ from PIL import Image, ImageFile
 import io
 
 from core.utils.utilities import Singleton
-from core.utils.win32.system_function import KEYEVENTF_EXTENDEDKEY, KEYEVENTF_KEYUP
 
 from winsdk.windows.media.control import (GlobalSystemMediaTransportControlsSessionManager as SessionManager,
                                           GlobalSystemMediaTransportControlsSession as Session,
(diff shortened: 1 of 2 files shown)


Video-Nomad:
Thanks for the PR! Works really well.
Two minor things I've noticed so far:
1. When I pause the video on YouTube and then close the tab it just gets stuck on that empty session. Is it possible to switch to the next active session if that happens?

https://github.com/user-attachments/assets/79f36ee8-cdc8-4311-8b77-75bfd4220bf8

2. Just some minor logging spam that's unnecessary unless it's DEBUG=True

image

Dusk-afk:
Thanks for the feedback! Glad to hear it’s working well. I've addressed both points:
1. Implemented a check to detect when the media info is empty and the current session is closed. If this condition is met, the session will automatically switch.
2. Added a condition to log only when `DEBUG` is enabled.

PR #181, opened 2025-03-29
docs: update installation instructions using python

trimclain:
Hey there,
I was trying to set up my development environment to develop some features for YASB and was following the instructions in the [wiki](https://github.com/amnweb/yasb/wiki/Installation#using-python), however I was constantly failing to install the requirements. After some research I found out, that the `winsdk` module requires Visual Studio installed together with 2 Individual Components, in my case `Windows 11 SDK` and `MSVC v143 - VS 2022 C++ x64/x86 build tools (Latest)`. Only after that did the `pip install -r requirements.txt` work. So I decided it would be nice to include that info in the wiki.
What are your thoughts on this? Did I miss anything?

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -19,6 +19,7 @@ winget install --id AmN.yasb
 
 ### Using Python
 - Install Python 3.12
+- Install Visual Studio with Individual Components `Windows SDK` and `MSVC build tools`
 - Install required Python Modules:
   - `pip install -r requirements.txt`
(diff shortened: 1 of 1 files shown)


amnweb:
Which Python version do you have installed?

trimclain:
Nevermind, I found my mistake. I have the latest python installed (3.13), and used [uv](https://github.com/astral-sh/uv) to create the venv using version 3.12. However, since I'm new to uv, I thought activating the venv would provide me the pip from the venv, which was not the case and I accidently ran pip install using 3.13.

PR #187, opened 2025-04-01
feat(volume): allow customizing muted volume text

trimclain:
Hey there!

Currently, whenever you mute your device, the `{level}` in the volume widget display the text `"mute"`. This value is hardocoded ([link](https://github.com/amnweb/yasb/blob/9004938cd28d16f7f274ffdb8b2262698d6df981/src/core/widgets/yasb/volume.py#L481)).

With this PR I want to allow users to customize that text. I did this by creating a new config option `mute_text`. This option is not required and defaults to `"mute"`, so there will be no breaking change.

Let me know what you think :)


amnweb:
Hi, just change `self._mute_level_volume` to `self._mute_text`.

PR #188, opened 2025-04-01
feat(cli): add `--silent` option

trimclain:
Hi,

First of all, huge thanks for creating the cli. I would never expect for a status bar on windows to have one and I'm amazed at how useful it is. I use `yasbc` mostly to `start`, `stop` and `reload` the bar in my glazewm config. Especially the `reload` command I find very useful, since chaining `taskkill /f /im yasb.exe` and `yasb.exe` on glazewm reload doesn't work as expected. Hovewer, since all commands have print messages, everytime they get called from glazewm, a terminal window spawns and dissappears. 

To remove that behavior I'd like to propose the `--silent` option. When passed, commands `start`, `stop` and `reload` will not print any messages, unless they come from the `stop_or_reload_application` method.

Usage:
```
yasbc start -s
```
```
yasbc reload --silent
```

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -20,6 +20,9 @@ The YASB CLI is a command line interface that allows you to interact with the YA
 - `--silent` - Disable print messages for `start`, `stop` and `reload`
 - `--version` - Show the YASB version.
 
+> **Note:**
+> You can use the `--silent` option with the `start`, `stop` and `reload` commands to prevent non-error messages from being displayed.
+
 ## Autostart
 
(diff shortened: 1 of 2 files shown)


amnweb:
Thanks for PR, but I think this is wrong, argument should be added at the end of command like `yasbc reload --silent` or `yasbc start --silent`

trimclain:
Done. Separate section in the Wiki or the Note is fine?

amnweb:
I'm fine with Note. Works great, just bump `YASB_CLI_VERSION` to `1.0.9`

PR #192, opened 2025-04-04
Update media.py

Forest-Terminal:
Added a callback that would allow for media to be played and paused by clicking the label.
### Check for bugs


amnweb:
Did you test this? If everything is working fine, please update the Wiki. Remove that code `#Check for bugs (written by a JS Dev)`, `#Update Wiki`, and so on. These are things we don't need in the main code.

Forest-Terminal:
I have updated the wiki and cleaned the code up. ***But*** I don't know how to check the code on my machine.

amnweb:
Added in this commit https://github.com/amnweb/yasb/commit/f9b5cb979cf4ce7bd9a9b1c1874b1fda303d16ce

PR #220, opened 2025-04-23
fix(notes): ensure widget is fully shown on one screen

Video-Nomad:
This fix prevents the widget from being split between monitors if the widget button is on the edge between two adjacent monitors.

Before:
lDRz5V6rm7

After:
N2rMrD2dVG


Video-Nomad:
Also for other widgets that use `PopupWidget`. I was focused on notes widget, but it will work for other things as well 😅

PR #223, opened 2025-04-24
feat(weather): add optional field 'lang'

ashinnotfound:
add a new optional field 'lang' to widget weather, so we can get the weather data in a specific language.
see [docs](https://www.weatherapi.com/docs/)
image


ashinnotfound:
> One of the reasons I didn't implement this is that half of the widget will be in English, while the other half will be in the language returned by the API. If we merge this, the next issue that will likely be opened is: `How do we translate the weather widget?`

true

Issue #239, opened 2025-05-02
Add hide_unsuported option to the battery widget


szbmrk:
### What is the request?

Add an option `hide_unsuported` to the battery widget so that the user can hide it if on the PC (similar to the brightness one).

### Why is it useful?

If a user has the same config on his PC and Laptop than on the PC it would be able to hide the battery widget.

PR #240, opened 2025-05-02
feat(battery): Add `hide_unsupported` option

szbmrk:
Feature requested in #239

Added new option to the Battery widget. The option `hide_unsupported` defaults to true. If the system cannot provide any battery info and the `hide_unsupported` option is set to true then it will hide the Battery widget.


amnweb:
Can I know why the timer needs to keep running all the time if the widget is hidden due to unsupported device ? I mean, if the device doesn't support have battery why do we need to check it constantly while YASB is running?

Maybe sometime like this
```
if self._hide_unsupported:
    self.hide()
    self.timer.stop()
    return
```

szbmrk:
You are right, i just fixed it.

PR #246, opened 2025-05-04
feat(glazewm_binding_mode): Add new widget called `GlazeWM Binding Mode`

szbmrk:
## Purpose
This widget shows the active binding mode for GlazeWM, so that the user can see if they are in for example Resize or Pause binding modes. With callbacks the user can disable the current binding mode as well.

## Concern
The GlazeWM websocket gives back a list of `bindingMode` when the binding mode is changed but as i tried it if i have a binding mode active and i want to activate another one the first one disables thus always resulting in no active binding modes or only one. And i haven't seen a config where anyone would use two or more binding modes at once. Therefore i implemented the widget to handle the first element of the given list. So i think it solves the concern but i can implement the widget to handle all the list elements as well if needed.

## Preview of widget
glazewm_binding_mode_widget


Video-Nomad:
Please rebase/squash some of those commits especially that merge one. Ideally leave only `docs` and `feat` should be left as they are the ones relevant.

szbmrk:
I'm new to contributing to open-source and using git like this, can you tell me how to do that please?

Video-Nomad:
If you are using github desktop just select those commits and right click - squash. It will allow you to edit the commit messages and in the end you will have only one commit with the new message.

If you are using git cli it will be interactive rebase `git rebase -i HEAD~4` (or however many commits you want to squash). It will open your default git editor and you can then replace "pick" with squash on the commits you want to squash together. If you are dealing with the merges you will have to use `git rebase -i --rebase-merges HEAD~4`. Same procedure after that.

I recommend you to use GitHub Desktop though if you are new to git.

Video-Nomad:
And after that when everything is good `git push -f`

Video-Nomad:
Thanks!

Video-Nomad:
Looks good to me. Tested and it seems to be working as expected.

Video-Nomad (approved):

Issue #137, opened 2025-02-13
GlazeWM-Workspaces option to define 'inactive_label' and 'active_label'


NoePfister:
### What is the request?

It is currently possible, to specify the text of ```populated_label``` and ```empty_label```, but no option to specify the label of the inactive/active workspaces.

### Why is it useful?

It would be nice to have this feature, so I can save some space and don't have to display the full name all the time.

### How could it be implemented?

I think, that it can be implemented the same as ```populated_label``` and ```empty_label```, but I don't know exactly how.

Video-Nomad:
Thanks for the suggestion! Will check it out when I have the time.

szbmrk:
@NoePfister I can implement this if it's not already implemented but i want to clarify something first.  If we have `populated_label`, `empty_label`, `active_label` and `inactive_label`, wouldn't the `populated_label` be unnecessery? I mean if i have for example 3 workspaces from which 1st is the active workspace it would get the `active_label` the second workspace is populated but inactive (i have focus on the 1st workspace) thus the label would be `inactive_label` and if the 3rd workspace is the same as the 2nd it would also get the `inactive_label` and every else workspace from 4 to 9 would get the `empty_label` because they are not used. Am i getting this wrong or we don't need the `populated_label` this way?

Video-Nomad:
Current styles:
`.active` - the workspace you are currently focused on (one per monitor).
`.populated` - the workspaces that have windows in them but are not currently focused.
`.empty` - the workspaces that are not currently focused and have no windows on them. Only available if `hide_empty_workspaces: false`

Current custom labels:
`populated_label` - is used when there's windows in the workspace, regardless if it's currently focused or not.
`empty_label` - is used when there's no windows in the workspace, regardless if it's currently focused or not.

From what I understand the request is to add at least one additional label that is `active` or currently focused and `populated` will be only used if it's not focused. Not sure what's the `inactive_label` though...

szbmrk:
Yes thank you for the clarification, so either way we don't need both `populated_label` and `inactive_label` at the same time. So the feature this issue should request is an additional `active_label` that can be used on the currently focused workspace.

szbmrk:
This issue can be closed now.

PR #249, opened 2025-05-04
feat(glazewm_workspaces): add `active label`

szbmrk:
Implements feature requested in #137 .


Video-Nomad:
Now that I look at the code I remember why I left only two states there (`populated` and `empty`).
Because it was either two like we have now, or four:
1. `active_populated`
2. `active_empty`
3. `populated`
4. `empty`
Makes sense I hope? And that's also what @NoePfister mentioned in #137 I think. Correct me if I'm wrong.

NoePfister:
Thank You👍

PR #252, opened 2025-05-05
feat(github): added notification dot option

Video-Nomad:
This option will show a small dot in a specified corner with specified color when there are unread notifications available. 
Can also have customizable x and y margin. 
It is drawn on the QLabel itself, so changing margin/padding for the `.icon` style will affect the dot position.

image

```yaml
notification_dot:
  enabled: True
  corner: "bottom_left" # Can be "top_left", "top_right", "bottom_left", "bottom_right"
  color: "red" # Can be hex color or string
  margin: [ 1, 1 ] # x and y margin for the dot
```

### DIFF
--- a/docs/widgets/(Widget)-Github.md
+++ b/docs/widgets/(Widget)-Github.md
@@ -1,21 +1,22 @@
 # Github Widget Options
 
-| Option           | Type     | Default                        | Description                                                                 |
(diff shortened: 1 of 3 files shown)


amnweb:
I like it. Thank you!👍

PR #257, opened 2025-05-08
feat(glazewm)!: add css classes `active_populated` and `active_empty`

trimclain:
PR #249 added labels `active_populated` and `active_empty`, however the CSS classes for them for not implemented. This PR adds missing classes, removes obsolete ones and updates the docs.

BREAKING CHANGE: replace `.glazewm-workspaces .ws-btn.active` with `.glazewm-workspaces .ws-btn.active_populated` in styles.css


Video-Nomad (approved):
Looks good. Will break styles for people but since we added extra state for label, it makes sense to add the style too.

PR #260, opened 2025-05-10
feat(komorebi): display opened app icons per workspace

Pitchaya-S:
### Summary:
Added a new button class `WorkspaceButtonWithIcons` to display opened app icons in each workspace, along with a new option:
```yaml
app_icons: 
  enabled: true
  enabled_active: true
  size: 12
  max_icons: 0
  hide_duplicates: false
  hide_floating: true
```

### Details:
- This new class uses `QFrame + QHBoxLayout` (similar to the stack widget button), applying styling via `.ws-btn .label` and `.ws-btn .icon`.
- The original button class is left completely untouched to preserve compatibility. The new class is only used when app icons are enabled, ensuring that existing themes won’t break after the update.
- Users who want to style the new icon-enabled buttons will need to update their custom CSS accordingly.

image

### DIFF
--- a/docs/widgets/(Widget)-Komorebi-Workspaces.md
+++ b/docs/widgets/(Widget)-Komorebi-Workspaces.md
@@ -8,6 +8,7 @@
 | `label_default_name`       | string  | `''`                     | The default name for workspaces.                                            |
 | `label_float_override`     | string  | `'Override Active'`                     | The label to display when Komorebi is Float Override Active. |
 | `toggle_workspace_layer`  | dict    | `{'enabled': false, 'tiling_label': 'Tiling', 'floating_label': 'Floating'}` | Controls toggling between tiling and floating layers.  |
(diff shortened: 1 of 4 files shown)


amnweb:
@Pitchaya-S This looks fine to me, but I found an issue with `enabled: true | false`, it doesn’t actually do anything?
Also, maybe you can replace all of this.
```
self._app_icons_enabled = app_icons['enabled']
self._app_icons_enabled_active = app_icons['enabled_active']
self._app_icons_size = app_icons['size']
self._app_icons_max = app_icons['max_icons']
self._app_icons_hide_duplicates = app_icons['hide_duplicates']
self._app_icons_hide_floating = app_icons['hide_floating']
```
with 
`self._workspace_app_icon = app_icons`

And than call it like 
`if self._workspace_app_icon['enabled']`
`if self._workspace_app_icon['hide_duplicates']`

PR #264, opened 2025-05-12
feat(css): add `.first-child` and `.last child` classes

NSPC911:
Adds support for [`:first-child`](https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child) and [`:last-child`](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child) only for widgets via classes

Usage:
```css
/* targets cava widgets that are the first child of their parent container*/
.cava-widget.first-child {
  background-color: red
}
/* targets cava widgets that are the last child of their parent container */
.cava-widget.last-child {
  background-color: green
}
```


NSPC911:
i hope the merge merged cleanly

PR #265, opened 2025-05-13
docs(whkd): fix typoo above preview image

szbmrk:
Fixes a typoo in the WHKD widget docs above the preview image.


amnweb:
Oh, I fixed this two minutes ago! :D

PR #305, opened 2025-05-28
Add right alignment option

dmperdue:
Added the option for the bar to be right aligned. For anyone else like me who might use the bar at smaller width percentages and wanted the option to left, center, or right align it.

- Updated `BAR_DEFAULTS` to use `alignment: ['left', 'center', 'right']`
- Modified `BAR_SCHEMA` to validate the new `alignment` field
- Refactored `bar_pos()` to calculate x-position for left/center/right
- Added example config snippet to README demonstrating right alignment

Manually tested on Windows 11 Home
Version 23H2 (OS Build 22631.5335)
Screenshot 2025-05-27 180833

### DIFF
--- a/src/config.yaml
+++ b/src/config.yaml
@@ -14,7 +14,7 @@ bars:
     class_name: "yasb-bar"
     alignment:
       position: "top"
-      center: false
+      alignment: "center"
     animation:
       enabled: true
       duration: 1000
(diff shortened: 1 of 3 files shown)


amnweb:
This PR would break the config for all users and themes in Gallery. To avoid that, we need to keep `center` under `alignment` in the validation file. While we can consider removing it in a future version. Additionally, `bar.py` should still include support for `center` as I've seen many themes on Reddit using `center: true`

So validation should be like 

```py
'alignment': {
    'position': 'top',
    'center': False, # Deprecated
    'alignment': 'center'
},
```

amnweb:
I've reverted the changes — there are way too many issues with bar position in this PR, especially when there are multiple bars on multiple screens

PR #310, opened 2025-05-29
feat(vscode): delete workspaces from history

szbmrk:
## purpose
On right click instead of opening delete the workspace from the history.

## concern
I mentioned it on the dc server, if i delete something and vscode is opened than it wouldn't reload the list, and if i open somefiles from the vscode recents tab than it would override the deletion. Everything else seems to work fine, i will get back to this problem later, but i think it can still be merged because it's not breaking anything.

### DIFF
--- a/docs/widgets/(Widget)-VSCode.md
+++ b/docs/widgets/(Widget)-VSCode.md
@@ -59,6 +59,10 @@ vscode:
 - **container_shadow:** Container shadow options.
 - **label_shadow:** Label shadow options.
 
+## Callbacks in the popup menu
+- **left click**: Opens the selected folder or file in Visual Studio Code.
+- **right click**: Deletes the selected folder or file from the opened history.
+
 ## Widget Style
 ```css
 .vscode-widget {}


szbmrk:
```	
/**
	 * Important: unlike other storage services in the renderer, the
	 * main process does not await the storage to be ready, rather
	 * storage is being initialized while a window opens to reduce
	 * pressure on startup.
	 *
	 * As such, any client wanting to access application storage from the
	 * main process needs to wait for `whenReady`, otherwise there is
	 * a chance that the service operates on an in-memory store that
	 * is not backed by any persistent DB.
 */
```
I found this in vscode's source code, so maybe this is our problem.

szbmrk:
they store like this:
```
	async removeRecentlyOpened(recentToRemove: URI[]): Promise<void> {
		const keep = (recent: IRecent) => {
			const uri = this.location(recent);
			for (const resourceToRemove of recentToRemove) {
				if (extUriBiasedIgnorePathCase.isEqual(resourceToRemove, uri)) {
					return false;
				}
			}

			return true;
		};

		const mru = await this.getRecentlyOpened();
		const workspaces = mru.workspaces.filter(keep);
		const files = mru.files.filter(keep);

		if (workspaces.length !== mru.workspaces.length || files.length !== mru.files.length) {
			await this.saveRecentlyOpened({ files, workspaces });
			this._onDidChangeRecentlyOpened.fire();

			// Schedule update to recent documents on macOS dock
			if (isMacintosh) {
				this.macOSRecentDocumentsUpdater.trigger(() => this.updateMacOSRecentDocuments());
			}
		}
	}
	private async saveRecentlyOpened(recent: IRecentlyOpened): Promise<void> {

		// Wait for global storage to be ready
		await this.applicationStorageMainService.whenReady;

		// Store in global storage (but do not sync since this is mainly local paths)
		this.applicationStorageMainService.store(WorkspacesHistoryMainService.RECENTLY_OPENED_STORAGE_KEY, JSON.stringify(toStoreData(recent)), StorageScope.APPLICATION, StorageTarget.MACHINE);
	}

```

szbmrk:
I will close this for now, will get back and do some research for this if i have time.

PR #311, opened 2025-05-31
docs(vscode): fix default callbacks

szbmrk:


amnweb:
Can you also include `label` and `label_alt` in example ?
image

PR #319, opened 2025-06-05
feat(komorebi_layout): tooltip support

AtifChy:
This PR just adds tooltip support for komorebi active layout.

### DIFF
--- a/src/core/validation/widgets/komorebi/active_layout.py
+++ b/src/core/validation/widgets/komorebi/active_layout.py
@@ -29,6 +29,7 @@ DEFAULTS = {
         'offset_left': 0,
         'show_layout_icons': True,
     },
+    'tooltip': True,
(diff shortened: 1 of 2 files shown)


amnweb:
I'm working on better tooltips, this native one looks really bad. So, I think this PR doesn't need to be merged because it will need to be changed again once I'm done.

https://github.com/user-attachments/assets/7218f996-0c7b-4a9f-b1a8-95db7454fd35

AtifChy:
Yeah ok, I just made this PR, because it was sometimes hard to understand which layout I am using if I set custom icons for layout.

amnweb:
I have just merged better tooltips into main, so if you still want to add a tooltip to this widget, please use...
```py
from core.utils.tooltip import set_tooltip
```

instead of
```py
self.setToolTip(layout_name)
```
You should use 
```py
set_tooltip(widget, layout_name)
```

PR #328, opened 2025-06-10
feat(glazewm): scroll through workspaces

szbmrk:
When hovering over the workspaces widget, scrolling with mouse wheel will scroll through the workspaces.

### DIFF
--- a/src/core/validation/widgets/glazewm/workspaces.py
+++ b/src/core/validation/widgets/glazewm/workspaces.py
@@ -48,7 +48,7 @@ VALIDATION_SCHEMA = {
     "hide_if_offline": {
         "type": "boolean",
         "default": DEFAULTS["hide_if_offline"],
-    }, 
+    },
     "container_padding": { 
         "type": "dict",
         "default": DEFAULTS["container_padding"],
--- a/src/core/widgets/glazewm/workspaces.py
+++ b/src/core/widgets/glazewm/workspaces.py
@@ -9,6 +9,7 @@ from PyQt6.QtCore import (
 )
 from PyQt6.QtGui import QCursor, QShowEvent
 from PyQt6.QtWidgets import QHBoxLayout, QLabel, QPushButton, QWidget
+from PyQt6.QtGui import QWheelEvent
 
 from core.utils.glazewm.client import GlazewmClient, Monitor
 from core.utils.utilities import add_shadow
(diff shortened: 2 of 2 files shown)


Video-Nomad (changes requested):
Works well on a single monitor setup. Just two observations:

1. If there are multiple monitors with workspaces like `1, 2, 3, 4` and `5, 6, 7, 8` it will scroll through all of them which is quite counterintuitive. You start scrolling on one screen expecting it to loop through that set of workspaces and it goes to the second screen. 
2. Another thing I noticed is that it also scrolls through empty workspaces by default, which is ok, but it would be great to have this as an option in the config.

I get that this complicates the implementation quite a bit and it will not be as elegant as it is right now, but it will improve the usefulness of this feature quite a lot.

szbmrk:
is there a way for me to test multiple monitor with one monitor?

Video-Nomad (commented):
Tested it. Works as expected with `scroll_through_empty: false`.
However, it starts scrolling through all the workspaces on both monitors again when `scroll_through_empty: true`

szbmrk:
when there is `scroll_through_empty: true` i use GlazeWM's in built `focus --prev-workspace` and so it does that, interesting. 

```python      
 for workspace in current_mon.workspaces:
            # Get or create workspace button if it's not present
            if (btn := self.workspaces.get(workspace.name)) is None:
                btn = self.workspaces[workspace.name] = GlazewmWorkspaceButton(
                    workspace.name,
                    self.glazewm_client,
                    display_name=workspace.display_name,
                    populated_label=self.populated_label,
                    empty_label=self.empty_label,
                    active_populated_label=self.active_populated_label,
                    active_empty_label=self.active_empty_label,
                )
                add_shadow(btn, self.btn_shadow)

            # Update workspace state
            btn.workspace_name = workspace.name
            btn.display_name = workspace.display_name
            btn.workspace_window_count = workspace.num_windows
            btn.is_displayed = workspace.is_displayed
```
this doesn't give us back the empty workspaces so i don't know how to work around that.

Video-Nomad:
Oh right, it's a problem with glaze, I forgot. It won't give you empty workspaces and you can't know what was in the user config from a simple IPC request... Same issue with `hide_empty_workspaces` as it won't show all of them until you actually click through them at least once.

szbmrk:
Maybe than i will get rid of the option and do it so it won't scroll through empty ones. Is that good?

Video-Nomad:
Yeah, I guess that's the best approach. At least until GlazeWM updates its IPC capabilities or we find a workaround for empty workspaces in general.

Video-Nomad (approved):
Works well 👍

PR #377, opened 2025-07-14
feat(clock): add per-hour icons support for ClockWidget

Isotech42:
This pull request introduces a new icons option in the YASB ClockWidget configuration. This option enables displaying a different icon (or text) depending on the current hour.

Details:

- Added the icons parameter in the YAML configuration, allowing a dictionary mapping each hour (clock_01, clock_02, …) to a Unicode character or a custom string.
- By default, times after 12:00 that are not explicitly defined in the icons revert to their corresponding icons in the interval clock_01 to clock_11.
- This feature enables richer visual customization of the clock widget, for example to indicate different moods or statuses depending on the hour.
- Documentation has been updated to explain how to use and configure this new option.
- No impact on existing widget functionality.

Thank you for your review and feedback!

### DIFF
--- a/docs/widgets/(Widget)-Clock.md
+++ b/docs/widgets/(Widget)-Clock.md
@@ -2,13 +2,14 @@
 
 | Option              | Type    | Default                                                                               | Description                                                                                                         |
 | ------------------- | ------- | ------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- |
-| `label`             | string  | `'\uf017 {%H:%M:%S}'`                                                                 | The format string for the clock. You can use placeholders like `{%H:%M:%S}` to dynamically insert time information. |
+| `label`             | string  | `'\uf017 {%H:%M:%S}'`                                                                 | The format string for the clock. You can use placeholders like `{%H:%M:%S}` or `{icon}` to dynamically insert time information. |
 | `label_alt`         | string  | `'\uf017 {%d-%m-%y %H:%M:%S}'`                                                        | The alternative format string for the clock. Useful for displaying additional time details.                         |
 | `class_name`        | string  | `""`                                                                                  | Additional CSS class name for the widget.                                    |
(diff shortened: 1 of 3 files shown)


amnweb:
That's a great idea, and we can expand on it by adding an extra class for the clock based on the hour. For example
```css
.clock-widget .icon.clock_23 {}
.clock-widget .label.clock_23 {}
```
so I can get a red clock at 23h :p

Isotech42:
It's actually a good idea that I hadn't even thought of. Now it's done! =D

amnweb:
Cool thanks, can you clean up the commit history a bit before the merge? To make this more readable in the git log.

Isotech42:
Oc ! I've squashed the commits into a single one for better readability. Let me know if there's anything else to adjust!

amnweb:
Thanks, looks good to me, but... I just thought, since this widget updates every 1 second, we can slightly optimize it. The idea is: what if we set `self.current_hour = None` inside `__init__` and in the `_update_label` we have 
`current_hour = f'{now.hour:02d}'`
Then, before doing any processing inside `update_label` we can compare 
```py
if self.current_hour != current_hour:
   self.current_hour = current_hour
   ...
```   
and continue with reloading the CSS and changing icons.

I know this isn’t a big issue, the process is fast anyway, but if we can optimize it without much effort, why not do it?

Isotech42:
Thanks for the suggestion! Don't worry, you're absolutely right, it's better to optimize what you can, especially if it's for like 3-4 lines of code.
I've just implemented it in the latest commit.
Now the ClockWidget skips icon and CSS updates when the time hasn't changed, which should avoid unnecessary QLabel refreshes during updates per second.

Let me know if there's anything else you'd like me to tweak or optimize, but I think this is good!

amnweb:
Looks good, thanks. There were some users who used two clock widgets on the same bar, one for time and another one just for date and calendar. And when you have it on 2 screens, this actually results in 4 updates per second. That's why I was thinking it would be better to have a simple check before processing any label update.

PR #387, opened 2025-07-25
feat(komorebi-stack): close windows with middle click on title

AtifChy:
Middle click on a stack window title will close that window gracefully.

### DIFF
--- a/src/core/utils/widgets/komorebi/client.py
+++ b/src/core/utils/widgets/komorebi/client.py
@@ -234,6 +234,12 @@ class KomorebiClient:
         except subprocess.SubprocessError:
             logging.exception("Failed to focus stack window")
 
+    def kill_window_by_pid(self, pid: int) -> None:
+        try:
+            subprocess.run(["taskkill", "/PID", str(pid)], check=True, capture_output=True, shell=True)
+        except subprocess.SubprocessError:
+            logging.exception("Failed to kill window")
+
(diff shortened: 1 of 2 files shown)


amnweb:
Do we really need to log every app when it is killed?
```py
logging.info(f"Killed window with PID {pid}")
```

AtifChy:
done

amnweb:
OK, thanks. But now I'm thinking about this before we merge it. First, I really don't like the `taskkill` idea. Also, this `kill_window_by_pid` function shouldn't be part of the komorebi client. I think it's much better to use something like what I use in the taskbar `win32gui.PostMessage(hwnd, win32con.WM_CLOSE, 0, 0)`, because it's much safer and can prevent data loss if you accidentally click on some apps.

Let me try to create a shared function to close apps so we can use it in all widgets where we need it. Also, I will remove the `mousePressEvent` from the stack widget and add callbacks there. After that, I will let you know so you can easily implement the rest of what you need.

amnweb:
@AtifChy 

```py
from core.utils.win32.utilities import close_application
```
and than 
```py
close_application(hwnd)
```

I'm not sure why @pitchaya-s set a `self.mousePressEvent = None`, this will prevent callbacks to this widget. I will talk with him on Discord, but you can implement middle click or maybe right click (not sure) to close the app.

Edit: btw do not forget to document this in the documentation

PR #397, opened 2025-07-30
Font family name mismatch

InCyberSpace:
Fixes the issue : [https://github.com/amnweb/yasb-themes/issues/168](url)


amnweb:
I think I’ve identified a major issue with this type of font check. For example, if you have DirectWrite enabled using `YASB_FONT_ENGINE='native'`, almost all themes will fail the font test.

_Example for the same font_
`YASB_FONT_ENGINE="native"` --> `JetBrainsMono Nerd Font Propo`
`YASB_FONT_ENGINE="gdi"`  --> `JetBrainsMono NFP`
`YASB_FONT_ENGINE="freetype"`  --> `JetBrainsMono NFP`


So, `set_font_engine()` in themes doesn't actually do anything if the theme defines `font-family: 'JetBrainsMono NFP'` and you're using `DirectWrite`

InCyberSpace:
> I think I’ve identified a major issue with this type of font check. For example, if you have DirectWrite enabled using `YASB_FONT_ENGINE='native'`, almost all themes will fail the font test.
> 
> _Example for the same font_ `YASB_FONT_ENGINE="native"` --> `JetBrainsMono Nerd Font Propo` `YASB_FONT_ENGINE="gdi"` --> `JetBrainsMono NFP` `YASB_FONT_ENGINE="freetype"` --> `JetBrainsMono NFP`
> 
> So, `set_font_engine()` in themes doesn't actually do anything if the theme defines `font-family: 'JetBrainsMono NFP'` and you're using `DirectWrite`

Then should we use an external library? 🤔

amnweb:
Which library?
If the theme you're trying to install include `font-family: 'JetBrainsMono NFP'`, but you're using `YASB_FONT_ENGINE="native"`, it won't work, because with the native font engine, you need to use the same font name as it appears in the system `font-family: 'JetBrainsMono Nerd Font Propo' `

 I'm not sure if this issue is specific to Nerd Fonts or if it affects all other fonts as well.

InCyberSpace:
> Which library? If the theme you're trying to install include `font-family: 'JetBrainsMono NFP'`, but you're using `YASB_FONT_ENGINE="native"`, it won't work, because with the native font engine, you need to use the same font name as it appears in the system `font-family: 'JetBrainsMono Nerd Font Propo' `
> 
> I'm not sure if this issue is specific to Nerd Fonts or if it affects all other fonts as well.

I made css_processor change fonts in the css to directwrite ones if the font engine is set to directwrite 
As for themes.py, I made it convert fonts to directwrite ones and check if whether fonts exist.
I included a prebuilt .dll, but one can build it oneself by running python build_font_family_util.py

amnweb:
Thank you for your effort, but this is not something I want to merge because it's over-complicated and messes up the codebase just for checking fonts. I'm sure this can be done more easily using only Python.
But to be honest, I don't currently care too much about this font checking and I'll probably remove it if I can't find a simple and effective solution.

InCyberSpace:
> Thank you for your effort, but this is not something I want to merge because it's over-complicated and messes up the codebase just for checking fonts. I'm sure this can be done more easily using only Python.
> But to be honest, I don't currently care too much about this font checking and I'll probably remove it if I can't find a simple and effective solution.

I think the main problem here is that correct fonts won't be loaded, which will happen even if you remove the font checking. Anyway thanks for taking the time 😃

PR #401, opened 2025-07-31
Add possibility to hide the systray icon with 'show_systray'

Draculagibus:
Let user define a parameter "show_systray"
The goal is to remove the YASB systray icon when the parameter is set to False

show_systray is True by default

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -49,6 +49,7 @@ Valid options are:
 | `watch_config`         | boolean    | `true`        | Reload bar when config is changed. |
 | `debug`      | boolean  | `false`   | Enable debug mode to see more logs |
 | `update_check`      | boolean  | `true`   | Enable automatic update check. This works only if the application is installed. |
+| `show_systray`      | boolean  | `true`   | Enable the YASB systray icon. |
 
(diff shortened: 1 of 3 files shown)


amnweb:
You can use `if config["show_systray"]:` since this option is always defined.
Also, could you please format the code for consistency? 

Additionally, please document the new options in the [https://github.com/amnweb/yasb/blob/main/docs/Configuration.md ](https://github.com/amnweb/yasb/blob/main/docs/Configuration.md) 

Lastly, make sure to write a clear, descriptive commit title and message, this helps keep the Git history clean and understandable.

Draculagibus:
Just to clarify, are you saying I **can't** use `if config["show_systray"]:` ?

I formatted the code correctly this time I hope.

I added the new option in the configuration.md in the **Status Bar Root Configuration** section

I understand your point, I tried to be more clear this time on the commit.
Thanks for your time and help !

amnweb:
I meant, instead of `if config.get('show_systray', True):` you can just use `if config['show_systray']:`, since the key is always defined.

amnweb:
Already added.

PR #408, opened 2025-08-05
Play sound on volume change

4zjata:
Made it for myself but i think some people might also like this addition.

### DIFF
--- a/src/core/widgets/yasb/volume.py
+++ b/src/core/widgets/yasb/volume.py
@@ -6,6 +6,8 @@ from ctypes import c_int as enum
 from ctypes.wintypes import BOOL, INT, LPCWSTR, WORD
 
 import comtypes
+import time
+import winsound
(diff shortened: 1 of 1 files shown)


amnweb:
I think this can be simplified:

```py
self.volume_slider.sliderReleased.connect(self._on_slider_released)
```

Then, instead of using `winsound`, we can use `ctypes`:

```py
def _on_slider_released(self):
    try:
        ctypes.windll.user32.MessageBeep(0)
    except Exception as e:
        logging.debug(f"Failed to play volume sound: {e}")
```


But to get this merged, we need an option to disable the sound, and it also needs to be documented.

airfortech:
would be awsome feature

amnweb:
> would be awsome feature

Already added https://github.com/amnweb/yasb/commit/c3797e64618dd227cb2293c13e1eb7be9bcf89b4

PR #457, opened 2025-09-14
Add rewrite feature to Komorebi stack widget

Demitrix2k:
- Added rewrite configuration to validation schema
- Implemented _rewrite_filter method in StackWidget
- Applied rewrite filter to window titles and process names
- Added regex pattern matching with case transformation support

I mainly just did this for personal preferences to use stack widget over active window and be able to just display program name without the ".exe" to it


amnweb:
You forgot to write the documentation. :)

PR #498, opened 2025-10-10
refactor(*): Used str.format and str.format_map in all widgets to insert values in labels

GhostOps77:
The earlier code used `str.replace` method to replace each placeholder in the labels with their values. This can be done by `str.format` and `str.format_map` methods. This will make the job easier to insert values.

So, instead of doing something like this
```py
label_values = {
    "icon": icon,
    "data": data
}

for key, value in label_values:
    formatted_text = part.replace('{' + key + '}', value) 
```
This can be done like this
```py
formatted_text = part.format(icon=icon, data=data)
```
or like this
```py
formatted_text = part.format_map({
    "icon": icon. "data": data
})
```
This is how it was done in this PR.


Video-Nomad:
Please format the code with `ruff` according to the `pyproject.toml`. You've changed the double quotes to single quotes in some files.

Also, there is a problem with `Libre Hardware Monitor` where it will briefly show the placeholders instead of an empty text while it's connecting on startup. Might be also the same for other widgets that update their labels this way. There's no reason to have this on startup. Just put an empty string as it was before.

https://github.com/user-attachments/assets/e6c22a09-e75e-4968-9bbf-f24f7c2d89ea

Video-Nomad:
Weather popup window is not working now. Please make sure that your refactor did not break anything else.
```less
  File "C:\Users\George\dev\windows\yasb\src\core\widgets\base.py", line 86, in _run_callback
    self.callbacks[callback_type](*callback_args)
  File "C:\Users\George\dev\windows\yasb\src\core\widgets\yasb\weather.py", line 139, in _toggle_card
    self._popup_card()
  File "C:\Users\George\dev\windows\yasb\src\core\widgets\yasb\weather.py", line 252, in _popup_card
    icon_url = self._weather_data[f"{{day{i}_icon}}"]
               ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
```

PR #521, opened 2025-10-24
feat(menu): Adds Menu widget with popup functionality

SavedByGrace251:
Introduces a new Menu widget that allows users to display a customizable menu with items in a popup window.

The menu can be configured with a label, icon, menu items, and various styling options such as blur, alignment, and animations.

It enables users to launch applications or execute commands directly from the bar.

desktop (1)
desktop (2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ https://github.com/user-attachments/assets/aab8d8e8-248f-46a1-919c-9b0601236ac1
 - **[Libre Hardware Monitor](https://github.com/amnweb/yasb/wiki/(Widget)-Libre-HW-Monitor)**: Connects to Libre Hardware Monitor to get sensor data.
 - **[Media](https://github.com/amnweb/yasb/wiki/(Widget)-Media)**: Displays media controls and information.
 - **[Memory](https://github.com/amnweb/yasb/wiki/(Widget)-Memory)**: Shows current memory usage.
+- **[Menu](https://github.com/amnweb/yasb/wiki/(Widget)-Menu)**: A customizable menu widget that displays items in a popup.
 - **[Microphone](https://github.com/amnweb/yasb/wiki/(Widget)-Microphone)**: Displays the current microphone status.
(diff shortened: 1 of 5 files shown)


amnweb:
Thanks for contributing! Since we already have a very similar widget Applications, I don’t think we need another one. But since your idea sounds good to me, maybe instead of creating a new widget, it would be better to extend the existing one with popup window support?

There’s already a defined app list, so we can reuse it for the popup if it’s enabled. However, this is a very old widget, and some parts need to be updated, especially the section where we run apps and also I think we can remove `ClickableLabel`. When apps are launched, they should be completely detached from YASB, similar to how it works in the taskbar.

https://github.com/amnweb/yasb/blob/65f1fd8fe27dc6ee3dd2a3ddf5b6e1eaf27a2b06/src/core/utils/widgets/taskbar/pin_manager.py#L641-L683

SavedByGrace251:
Oh I didn't even think about extending the applications widget. I'll take a look at extending it and report back!

amnweb:
> Oh I didn't even think about extending the applications widget. I'll take a look at extending it and report back!

Just make sure you don’t break the current users’ configurations if you decide to work on this widget, many themes still use it.

SavedByGrace251:
Totally, my thought is a new optional param `use_popup` that would then look for a label and put item into a pop-up menu instead of a regular list.

CodeRedPaul:
Hey @SavedByGrace251 

I've been stealing your code!

I needed a popup menu to extend an existing widget, but python has never been my thing - I'm learning now but I've been a .NET developer for over 25 years and I know nothing about python...

Wondering if you wouldn't mind giving me a pointer (pun intended - you see why soon!). 

I've taken a bunch of your code and retrofitted it into a different widget but I've hit the current limit of my understanding and don't know how to move things forward. It shouldn't take too much time - I'm just confused as to the mouse pointer events and why the popup control and original widget seem to be getting confused on the mouse click.

Would you mind giving me a quick hand, please? If you're ok with that then I'll add you as a contributor to my fork.

Thanks mate.... And hope you have a good Lent. 😉

Cheers.

[EDIT - 2023-03-04: I got it sorted, thank you!]

Issue #540, opened 2025-11-05
(Widget)-GlazeWM-Workspaces - Add an Optional Distinction Between an Active Workspace and a Focused Workspace


EmilBerbakov:
### What is the request?

Adding optional customization to the GlazeWM-Workspaces widget that would allow you to have visual distinction between an empty or populated active workspace and its focused workspace equivalent, while also making sure that any current customization that was made assuming an active workspace would also be focused won't be affected.

### Why is it useful?

When using GlazeWM on a default multi-monitor setup, each monitor has a workspace. Each workspace is considered active, even if it is not focused (i.e. you are focused on a window/tile within the workspace/monitor). This means that swapping focus between two workspaces that are either both populated or both empty produces no visual feedback. This enhancement would allow that, should the user decide to implement it.

### How could it be implemented?

I have a fork where I implemented this. It's mostly done, except I either need to alter some of my logic within GlazewmWorkspaceButtonWithIcons, or I'm having a CSS skill issue (probably a combination of both, if I'm honest, lol). I cannot get the .label customization to behave in an expected way, as of yet.

The basic ideas for implementation in a bullet-point list are:
### In the yaml options
- add focused_populated_label and focused_empty_label. Both of these would function exactly like their active_\*\_label equivalents, but if you don't explicitly set them, they take on the values of their active_\*\_label counterparts, to preserve current functionality
- add a nullable property called enabled_focused that takes on the value of enabled_active if not explicitly set
### In the widget code
- add two new statuses to the WorkspaceStatus enum: FOCUSED_EMPTY and FOCUSED_POPULATED
- add a new property to the two GlazewmWorkspaceButton classes called is_focused that listens to the focus property of the Workspace class
- when applying status-based classes to the ws-btn with a FOCUSED_EMPTY or FOCUSED_POPULATED status, apply the equivalent active class first so that you have to opt into making the focused statuses visually distinct from their active counterparts
  - this means that you have to implement the focused classes after their active counterparts in the css file in order for them to have an effect. It also means any property that the active class implemented that you don't want in the focused class needs to be unset in the focused class. I don't super like this, but I can't think of a way to do this without introducing a breaking change to those who don't want to opt into this extra customization.

If there's interest in adding this, I'd be happy to make a PR out of my fork. :)

### Additional context

Here's an example image set of my implementation in action. Below them are the customization I made in the yaml and CSS, but the tl;dr is:
- red background with black text: focused and populated
- blue background with white text: active and populated
[...]

PR #550, opened 2025-11-10
Glazewm Workspaces - Add Optional Distinction Between Active and Focused Workspaces

EmilBerbakov:
Implements the request asked for here [(Widget)-GlazeWM-Workspaces - Add an Optional Distinction Between an Active Workspace and a Focused Workspace](https://github.com/amnweb/yasb/issues/540).

This PR adds the following:

### Widget Code
- added two new statuses: FOCUSED_EMPTY and FOCUSED_POPULATED
- added a refresh_widget_style call for the text_label of GlazewmWorkspaceButtonWithIcons
  - this allows the ability to add targeted, status-specific styling to the .label class of the button when app icons are enabled

### yaml options:
- focused_populated_label and focused_empty_label
  - These function like their active counterparts. If not set, they will take on the values of their active counterparts to maintain current functionality
- enabled_focused
  - This enables app icons for focused workspaces. It is nullable, and if not set, takes on the value of active_focused to maintain current functionality

### CSS customization:
- focused_populated and focused_empty
  - Whenever a workspace is determined  to be FOCUSED_POPULATED or FOCUSED_EMPTY, both the relevant active class is applied, followed by the relevant focused class. This allows for CSS customization for focused statuses to be entirely optional, and all previous stylings will be unaffected.
    - This also means that, in order to opt in, one must define the focused styles _after_ their respective active stylings. This quirk has been added to the documentation

### DIFF
--- a/src/core/widgets/glazewm/workspaces.py
+++ b/src/core/widgets/glazewm/workspaces.py
@@ -216,6 +216,8 @@ class GlazewmWorkspaceButtonWithIcons(QFrame):
             button_class += self.status.value
         self.setProperty("class", button_class)
         refresh_widget_style(self)
+        # Even though the label class name, we still need to run this on the label to catch any different stylings we want to do when the status changes
+        refresh_widget_style(self.text_label)
 
     def _update_label(self):
         replacements = {


Video-Nomad (approved):
Thanks for the PR! The code looks good to me, and I can't find any issues with the styling or backwards compatibility. Without changing anything my old config looks the same and the new stuff works too if I add it 👍

Just a small note (for myself mostly not related to this PR): there is a small styling bug with icons enabled and empty (not active) workspaces not having the correct style from `.glazewm-workspaces .ws-btn.empty`. In GlazeWM workspaces should be set to `keep_alive: true`. If you want to check this out and fix it in this PR that would be great but it's fine to leave that, I can fix it later.

Another note on commit history - please clean it up a bit. It's fine to have multiple commits in a PR but please check other commits and how they are written and styled. Maybe rebase and squash some of them if they are redundant.

EmilBerbakov:
I have updated the commit history to be a lot cleaner.

For the empty styling, are you saying that a note needs to be added to the documentation on how a workspace needs to have keep_alive set to true if they want the empty styling to be applied to said workspace, or are you saying there's a way to dynamically set that property for a workspace through a command?

Video-Nomad:
Sorry, nevemind, it was my mistake with the styling. Just re-tested it again and it works as expected. `.glazewm-workspaces .ws-btn.empty` should be `.glazewm-workspaces .ws-btn.empty .label` for when the icons are enabled. Forgot this detail.

So, this PR is good to go then, at least on my end. I can't merge stuff so it's up to @amnweb.

PR #561, opened 2025-11-20
feat(custom): add tooltip support to custom widget

terakael:
Enables custom widgets to display tooltips, with the ability to use the same data object that the label has access to.

### DIFF
--- a/.github/workflows/pr-check.yaml
+++ b/.github/workflows/pr-check.yaml
@@ -10,7 +10,6 @@ on:
 
 permissions:
   contents: read
-  pull-requests: write
 
 env:
   PYTHON_VERSION: '3.14'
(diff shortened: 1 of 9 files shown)


amnweb:
This can be in one line
```py
if not self._tooltip:
    return

if not self._exec_data:
    return
```
like 
```py
if not self._tooltip or not self._exec_data:
```

Remove `import json` line 227


Also, if someone has HTML in JSON like I do, it will show in the tooltip
Screenshot 2025-11-21 172127
Screenshot 2025-11-21 172332


Maybe we can remove any HTML tags from the tooltip text ? Something like 
```py
tooltip_text = re.sub(r'<[^>]+>', '', tooltip_text)
```
Screenshot 2025-11-21 173159


**Edit:** Also, please clean up the commit history so we can have a single clear commit about this

terakael:
Thanks for the review!  I've made the changes you requested.

For the json showing the raw html - I feel like if you used the tooltip_label feature like:

```
tooltip_label: "{data[absoluteChange]}<br>{data[lastPrice]}"
```

you'd get your expected formatting (similar to how the Weather widget is working).  This also leaves it up to the one configuring their widget to choose whether they want the full raw json or to have some formatting.

What do you think?

PR #595, opened 2025-12-15
feat(ai-chat): add toggle-widget support, auto focus input option and default model selection

Rejdukien:
* Adds support for `yasbc toggle-widget ai_chat`
* Auto focuses text input field when widget is opened
* Add a default model selection, also configured via yaml, `default: true` added to the respective model. Only affects initial initialization, if the widget is opened, model/provider changed, closed and opened back up the default model doesn't affect it.  
Also, should more than one model be declared as default, first model is selected as default.

Just started using yasb, figured I'd implement the changes I'd like to see myself. Should make it easier to use the widget via hotkeys.

Tested behavior with "old" yaml before the change, no breaking changes from what I could see.

### DIFF
--- a/docs/widgets/(Widget)-Ai-Chat.md
+++ b/docs/widgets/(Widget)-Ai-Chat.md
@@ -58,6 +58,7 @@ ai_chat:
           label: "GPT3.5 Turbo"
         - name: "gpt-4"
           label: "GPT4"
+          default: true
           temperature: 0.3
           top_p: 0.95
           max_tokens: 4096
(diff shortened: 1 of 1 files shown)


Rejdukien:
~~I went ahead and made it possible to toggle specific instances of the widget by adding a widget_config_name attribute to the base widget, which gets populated by the widget builder. This name can be referenced when calling toggle-widget - at least for this widget, haven't touched the `_handle_widget_cli` method of other ones. Don't think it'd make much sense to have multiple launchpads, wallpaper managers or power menu widgets.
Figured it's not a bad idea, since technically it's possible to add multiple widgets of the same type.~~

Edit: Food for thought later, removed from this PR.

amnweb:
Interesting, Ai makes Ai widget xD

Rejdukien:
Ayy, not gonna lie, for smaller changes in an unfamiliar codebase it's pretty decent - as long as you verify afterwards. 😛

amnweb:
I will check all this when I have some free time, probably this weekend

Rejdukien:
Okay, should be everything now.
* Simplified it to just always focus when showing the widget, additionally it also moves to cursor to the end of the input field.
* Changes to widget-builder and related changes have been removed from the PR.
* auto_focus_input removed from schema validation and docs.

PR #615, opened 2025-12-30
feat(ai-chat): Attachment support

Rejdukien:
This PR adds support for attachments to the ai-chat widget.
Attachments can either be added via file-picker dialog or via clipboard pasting in the input field.
Supports text files (appended to the api payload as `type: text`) or images (included as `type: image_url`, base64 encoded). Binary files (that can't be utf-8 decoded) get rejected as unsupported file type.

There's two parameters to limit the max size per attachment: Text files get truncated, image files get compressed if they exceed that limit.

Supposedly OpenAIs chat completion endpoint should also allow pdf files, but in my testing I've noticed that the provider/model I use doesn't support it, so I've decided to not include untested changes.

This PR also includes a minor change to the popup / event-handling logic to properly support the file picker dialog (without the chat popup closing on me...)

image

Image shows one text file and two images attached to the first message sent to the model; two more files are staged for the next message. Staged attachments can be removed at this point via the deletion button next to each one.

Let me know if there's anything you don't like and want changed.
Happy new year!

### DIFF
--- a/docs/widgets/(Widget)-Ai-Chat.md
+++ b/docs/widgets/(Widget)-Ai-Chat.md
@@ -32,6 +32,7 @@ ai_chat:
       offset_left: 0
       offset_top: 6
     icons:
+      attach: "\uf0c6"
       send: "\uf1d8"
       stop: "\uf04d"
       clear: "\uf1f8"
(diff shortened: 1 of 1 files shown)


amnweb:
Is it possible to paste an image from the clipboard? :)

Rejdukien:
> Is it possible to paste an image from the clipboard? :)

Yup!
image

See method _handle_paste_mime for the implementation of that.

PR #625, opened 2026-01-03
fix(docs): incorrect paths to previews

notTamion:


amnweb:
I think this is wrong. The GitHub action will release the Wiki from these docs, and probably all images will be missing with paths like `../`, but I will test it later.

notTamion:
ah wasn't aware the docs there get published. right now when looking at them through the files it will result in them not being found.

>  but I will test it later.

would be awesome, thanks! Otherwise feel free to close

amnweb:
The only thing I found that works is using an absolute/RAW URL, like `AI Chat YASB Widget`.

I don’t think this is a big deal, the Wiki is already published at all times for the latest and dev releases on https://docs.yasb.dev/latest

notTamion:
will close this then. thanks tho

PR #656, opened 2026-01-25
refactor(ai-chat): image processing utilities for attachments

Rejdukien:
Not completely finished yet - I'd still like to try adding background threading for image processing to not freeze the UI, but have been having some issues with that so I left it out.

Two things here.
I've tried to move most of the image processing stuff out into its own helper.
In the helper, we're raising the allocation limit of QTs image reader, then scaling the input image down to more sensible limits, before trying to compress the file to the output image size limit set by the user in their model configuration.

This should prevent us from running in the QImage allocation limit if unreasonable input images are chosen and also improve token efficiency by quite a bit. It probably wouldn't be a bad idea to also lower the MAX_IMAGE_DIMENSION even more to what was stated in OpenAIs documentation, for those people that use per-token billing.
Iterative quality-decreasing was left in, we're also trying to save it in the original format first, before falling back to jpeg (if it doesn't fit within user image size limits). Probably important for those formats that support multiple frames, such as gif or apng, though I honestly don't know if the vision part of most LLMs even supports that.

If this is overkill, you could also just take the part from `image_helper.py` line 50 to line 92 and plug that into the existing `ai_chat.py` within `_compress_image_if_needed`.

### DIFF
--- a/src/core/utils/widgets/ai_chat/image_helper.py
+++ b/src/core/utils/widgets/ai_chat/image_helper.py
@@ -1,0 +1,213 @@
+"""Image processing utilities for AI chat widget."""
+
+import logging
+from typing import Any
+
+from PyQt6.QtCore import QBuffer, QIODevice, QSize
+from PyQt6.QtGui import QImage, QImageReader
+
+FORMAT_TO_MIME = {
+    "PNG": "image/png",
+    "JPEG": "image/jpeg",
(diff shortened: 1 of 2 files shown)


Rejdukien:
Closing due to 7b3c66783be5edd5f38e2af6fb3df96ff8d58981

PR #670, opened 2026-02-02
feat(battery): add option to set unlimited time remaining icon/text

stylebending:
Hi! This PR addresses the request in [Discussion #663](https://github.com/amnweb/yasb/discussions/663).

I’ve added a new configurable icon, `icon_unlimited`, to the battery widget's `status_icons`. This allows users to choose between the default "unlimited" text or the infinity icon "∞" (or any other string/icon) for a cleaner taskbar look.

### Changes:
* Added `icon_unlimited` to `DEFAULTS` and `VALIDATION_SCHEMA`.
* Updated `_get_time_remaining` to pull the value dynamically from the user's configuration.
* Updated the battery widget docs to explain the option.
* Default behavior remains as "unlimited" to maintain backward compatibility.

### DIFF
--- a/docs/widgets/(Widget)-Battery.md
+++ b/docs/widgets/(Widget)-Battery.md
@@ -7,10 +7,11 @@
 | `class_name`            | string  | `""`                                         | Additional CSS class name for the widget.                                    |
 | `update_interval`       | integer | `5000`                                       | The interval in milliseconds to update the widget.                          |
 | `time_remaining_natural`| boolean | `False`                                      | Whether to display the remaining time in a natural format.                  |
+| `time_remaining_unlimited_icon`| string | `unlimited`                                      | Which string to use to display unlimited time remaining.                  |
 | `hide_unsupported`| boolean | `True`                                      | Whether to hide the widget if the current system does not have battery info.                  |
 | `charging_options`      | dict    | `{icon_format: '{charging_icon}', blink_charging_icon: True, blink_interval: 500}` | Options for charging state display.                                         |
 | `status_thresholds`     | dict    | `{critical: 10, low: 25, medium: 75, high: 95, full: 100}` | Thresholds for different battery statuses.                                  |
-| `status_icons`          | dict    | `{icon_charging: '\uf0e7', icon_critical: '\uf244', icon_low: '\uf243', icon_medium: '\uf242', icon_high: '\uf241', icon_full: '\uf240', icon_unlimited: '\u221e'}` | Icons for different battery statuses.                                       |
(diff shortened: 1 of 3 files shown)


Video-Nomad:
Hi! Thanks for the PR. We recently moved from Cerberus to Pydantic for validation files. Please check the new battery validation schema format.

stylebending:
> Hi! Thanks for the PR. We recently moved from Cerberus to Pydantic for validation files. Please check the new battery validation schema format.

Thanks for pointing that out! I've updated the battery validation schema and the widget's `_get_time_remaining` function to match the new project standard. This is now ready for another review.

Video-Nomad (commented):
This works as described, however my problem with this change is that we are adding an icon to the `status_icons` but it's not a status icon `{icon}` . It's a separate `{time_remaining}`. It does not have specific styling like other icons. It's more similar to `charging_options` to be honest, so it might be a better idea to expand this and make something like:
```
time_remaining_options:
  time_format: <- Standard python time formatting string like "%H:%M:%S"
  unlimited_icon: <- Unlimited icon
```
We can't really move current `time_remaining_natural` as it will break configurations, but `time_format` can be overridden by it if it's `true`. Ideally it would go to the same `time_remaining_options`.
@amnweb What do you think about this?

amnweb:
@Video-Nomad I'm fine with any option you choose since I only have a battery device when I mock one D
Not sure if you really need `%S` there, but if this report shows the correct remaining time then it's ok.

Video-Nomad:
@stylebending So I guess to not overcomplicate this too much and since we already have `time_remaining_natural` I think just renaming the `icon_unlimited` -> `time_remaining_unlimited_icon` and putting it right under `time_remaining_natural` would make more sense. Just to separate it from `status_icons`. And leave it at that.

```yaml
battery:
  type: "yasb.battery.BatteryWidget"
  options:
    time_remaining_natural: False
    time_remaining_unlimited_icon: "\u221e"
```

stylebending:
@Video-Nomad Thanks for taking a look at this again! I've made these changes:
 - Renamed `icon_unlimited` to `time_remaining_unlimited_icon`.
 - Moved it from `status_icons` to right under `time_remaining_natural`.
 - Edited the widget's documentation to represent this correctly.
 
 I believe this is ready now, if there's anything else that needs to be changed, I would love to hear it.

Video-Nomad (approved):

Video-Nomad:
Can you cleanup the commit history a bit @stylebending . Just rebase and squash everything. There's like 5 lines changed, no need to have 4 commits 😅

PR #684, opened 2026-02-06
feat(media): add open_media_source callback option

stylebending:
### Description
This PR addresses [Discussion #683](https://github.com/amnweb/yasb/discussions/683) to add a callback option to the Media Widget to open/activate the application currently playing media by clicking the widget.

### Changes
* `docs/widgets/(Widget)-Media.md`: Added `open_media_source` to callback documentation.
* `src/core/utils/win32/aumid.py`: Added `activate_app_by_aumid(aumid: str)` utility function. This uses `EnumWindows` to find the window handle (`HWND`) associated with a given App User Model ID (`AUMID`) and brings it to the foreground.
* `src/core/widgets/yasb/media.py`: 
    * Imported `activate_app_by_aumid`.
    * Added `_open_media_source` callback method.
    * Registered `open_media_source` as a valid callback string.

### Configuration
Users can now configure this action in their `config.yaml`:

```yaml
widgets:
  media:
    options:
      callbacks:
        on_left: "toggle_media_menu"
        on_middle: "toggle_play_pause"
        on_right: "open_media_source"
```

### Rationale
While `source_apps.py` exists for mapping AUMIDs to display names, it does not provide window handles. The new `activate_app_by_aumid` function fills this gap by dynamically finding the window owned by the media source's `AUMID`, for activating any media application.

If any tweaks are needed, or if I'm taking the wrong approach, I would love to hear the feedback to help out!

### DIFF
--- a/docs/widgets/(Widget)-Media.md
+++ b/docs/widgets/(Widget)-Media.md
@@ -169,6 +169,7 @@ media:
 - `toggle_label`: Toggles the visibility of the label.
 - `toggle_play_pause`: Toggles between play and pause states.
 - `toggle_media_menu`: Toggles the visibility of the media menu popup.
+- `open_media_source`: Opens the the source that is playing the media.
 - `do_nothing`: A placeholder callback that does nothing when triggered.
 
 ## Description of Options
--- a/src/core/utils/win32/aumid.py
+++ b/src/core/utils/win32/aumid.py
@@ -66,13 +66,25 @@ PKEY_AppUserModel_ID = PROPERTYKEY(GUID("9F4C2855-9F79-4B39-A8D0-E1D42DE1D5F3"),
 
 class IPropertyStoreVtbl(ctypes.Structure):
     _fields_ = [
-        ("QueryInterface", WINFUNCTYPE(ctypes.c_long, c_void_p, POINTER(GUID), POINTER(c_void_p))),
+        (
+            "QueryInterface",
+            WINFUNCTYPE(ctypes.c_long, c_void_p, POINTER(GUID), POINTER(c_void_p)),
+        ),
         ("AddRef", WINFUNCTYPE(ctypes.c_ulong, c_void_p)),
         ("Release", WINFUNCTYPE(ctypes.c_ulong, c_void_p)),
         ("GetCount", WINFUNCTYPE(ctypes.c_long, c_void_p, POINTER(ctypes.c_uint))),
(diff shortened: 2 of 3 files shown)


Video-Nomad:
Thanks for the PR! Quickly testing it and it only works for my browser as an audio source. Spotify app and YouTube Music (third party) app does not react to this callback.

Edit: it works when the application is on another workspace (komorebi or glazewm) but not when it's minimized to taskbar/tray. I assumed this callback will also open the app, but it just changes the focus for me.

stylebending:
> it works when the application is on another workspace (komorebi or glazewm) but not when it's minimized to taskbar/tray. I assumed this callback will also open the app, but it just changes the focus for me.

Thanks for testing and the feedback, much appreciated! I've pushed a fix to handle minimized windows and also to focus the source window to make sure it is able to receive input immediately. This should be ready for another review.

Video-Nomad:
So tried testing multiple apps and some apps are working properly now like web browsers (Edge, Vivaldi), Spotify, Windows Media Player (the modern one) and various PWAs I have.
What's not working for some reason is the unofficial [YouTube Music](https://github.com/pear-devs/pear-desktop) app, [MPC-HC](https://github.com/clsid2/mpc-hc) fork and telegram when it's playing a voice message for example, even though they report their media information to the media widget...
Might be something very specific about those, but other than that seems like the PR works as expected.

amnweb:
https://github.com/amnweb/yasb/blob/15857a72c191c58498fa65d07e4ddf10aa83512c/src/core/utils/win32/window_actions.py#L77-L105

@stylebending what about these functions? Or do you still need to define **restore if minimized** inside AUMID?

stylebending:
Updated! I'm now using `restore_window` and `force_foreground_focus` from `window_actions.py` to handle the restoration logic, thanks for the suggestion @amnweb! Added a fallback that checks process names (via `source_apps.py` mapping) if the exact AUMID match fails. This should fix activation for apps like unofficial YouTube Music clients and MPC-HC, I don't have them myself it would be awesome if whenever you have the spare time you could test those again @Video-Nomad Thanks again for helping out!

Tested with GlazeWM, Brave, Chrome, Firefox, Spotify, Windows Media Player, seems to be working fine for all of them.

Video-Nomad (approved):
YouTube Music client works, can confirm. MPC-HC does not but that's not a problem. It's a weird one. Also other apps mentioned work just fine.
I think this one is good 👍

PR #778, opened 2026-03-12
feat(media): add max size mode

oOOoutsiderOOo:
Added a new option to set a fixed width for the media widget. Using `size_mode: "auto"` it behaves as normal, following the label width. When setting `size_mode: "max"`, it renders at max_field_size width regardless of the actual label width.


Video-Nomad:
Thanks for the PR!

Is there a way to make this work with stylesheet alone?
This will set the fixed width of the label, but now that I test it it does not behave reliably, so I get where this PR comes from.

```css
.media-widget .label {
    min-width: 400px;
    max-width: 400px;
}
```

Would be great if we could control that via stylesheet alone.

PR #951, opened 2026-06-05
feat: add Claude usage widget

ManaphatDev:
Adds a widget showing Claude Code subscription usage (5-hour and 7-day limits) on the bar, with a popup card of progress bars and reset countdowns. Reads the existing Claude Code OAuth token from ~/.claude/.credentials.json and caches results in app_data_path with a TTL (429-safe). Includes a docs page and regenerated schema.json.

### DIFF
--- a/docs/_Sidebar.md
+++ b/docs/_Sidebar.md
@@ -16,6 +16,7 @@
     - [Bluetooth](./(Widget)-Bluetooth)
     - [Brightness](./(Widget)-Brightness)
     - [Cava](./(Widget)-Cava)
+    - [Claude Usage](./(Widget)-Claude-Usage)
     - [Copilot](./(Widget)-Copilot)
     - [CPU](./(Widget)-CPU)
     - [Clock](./(Widget)-Clock)
--- a/docs/widgets/(Widget)-Claude-Usage.md
+++ b/docs/widgets/(Widget)-Claude-Usage.md
@@ -1,0 +1,164 @@
+# Claude Usage Widget Options
+
+Shows your Claude (Claude Code) subscription usage on the bar — the 5-hour and 7-day
+rolling limit utilization — with a popup menu that displays both windows as progress bars
+together with their reset times (a countdown plus the exact reset timestamp).
+
+The data is read from the same OAuth credentials the Claude Code CLI uses
+(`~/.claude/.credentials.json`) and fetched from Anthropic's usage endpoint. No API key or
+extra configuration is required as long as you are signed in to Claude Code.
+
+| Option            | Type    | Default | Description |
+|-------------------|---------|---------|-------------|
+| `label`           | string  | `'Claude {five_hour}%'` | The format string for the label. Supports the placeholders below. |
+| `label_alt`       | string  | `'Claude {seven_day}%'` | The alternative format string, toggled by the `toggle_label` callback. |
+| `update_interval` | integer | `60` | How often the label and reset countdown are refreshed, in seconds. Must be between 30 and 3600. |
+| `cache_ttl`       | integer | `120` | How long (seconds) a fetched result is cached on disk before the endpoint is queried again. The endpoint is rate-limited, so keep this at a sane value. |
+| `tooltip`         | boolean | `true` | Whether to show a summary tooltip on hover. |
+| `callbacks`       | dict    | `{'on_left': 'toggle_menu', 'on_middle': 'do_nothing', 'on_right': 'toggle_label'}` | Mouse-click callbacks. |
+| `menu`            | dict    | `{'blur': true, 'round_corners': true, 'round_corners_type': 'normal', 'border_color': 'System', 'alignment': 'right', 'direction': 'down', 'offset_top': 6, 'offset_left': 0}` | Popup menu settings. |
+
+## Placeholders
+
+The label is plain text by default. You can prepend a Nerd Font glyph in a `<span>` if you
+want an icon (e.g. `<span>\U000f06a9</span> {five_hour}%`). The following placeholders can be
+used in `label` / `label_alt`:
+
+- `{five_hour}` — 5-hour window utilization (percent, `--` when unavailable).
+- `{seven_day}` — 7-day window utilization (percent, `--` when unavailable).
+- `{five_hour_reset}` — time until the 5-hour window resets. Shown as a countdown when under a
+  day away (e.g. `4h 27m`), otherwise as a local weekday + time (e.g. `Sat 6:00 AM`).
+- `{seven_day_reset}` — time until the 7-day window resets (e.g. `Sat 6:00 AM`).
+
+```yaml
+claude_usage:
+  type: "yasb.claude_usage.ClaudeUsageWidget"
+  options:
+    label: "Claude {five_hour}% · {five_hour_reset}"
+    label_alt: "Claude 7d {seven_day}% · {seven_day_reset}"
+    update_interval: 60
+    cache_ttl: 120
+    callbacks:
+      on_left: "toggle_menu"    # open the usage menu
+      on_middle: "do_nothing"
+      on_right: "toggle_label"  # switch the bar text between 5h and 7d
+    menu:
+      blur: true
+      round_corners: true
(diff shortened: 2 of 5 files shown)


amnweb:
1. `schema.json` is an auto generated file and it is not allowed to be changed.

2. Separate the service from the widget. Create something like `claude_api.py` in `/src/core/widgets/services/claude_usage/` and use this file to share data between multiple widgets. Example:

```text
       +-----------------------+
       |       Claude API      |
       +-----------+-----------+
                   |
     +-------------+-------------+
     |             |             |
+----+-----+ +-----+----+ +-----+----+
| Widget 1 | | Widget 2 | | Widget 3 | 
+----------+ +----------+ +----------+
```

3. You do not need to use a hardcoded SVG for the icon in the label. This is not the standard way YASB works, as you can easily link an image in a label using an `` tag. So:
```
icon_size: 14
icon_color: "#D97757"
```
you probably do not need this.

4. Since this PR was submitted by AI and everything looks wrong, try prompting it first to understand the codebase and how the widget service works. Even better if you can understand it properly so you can fix everything they made.

5. Use `menu:` instead of `card:` for the popup.

ManaphatDev:
Thanks for the detailed review I went back through the widget/service architecture properly this time and pushed the changes. Addressed point by point:

1. schema.json reverted to upstream, so it's no longer part of this PR.

2. Service / widget split moved all the data fetching into a shared service at src/core/widgets/services/claude_usage/claude_api.py. ClaudeUsageService is a reference-counted singleton keyed by (update_interval, cache_ttl) that owns the timer + worker thread and emits a data_ready signal, so multiple widgets share a single poller and a single on-disk cache following the server_monitor service pattern. The widget is now just a view that subscribes to the signal.

3. Icon - dropped the hardcoded SVG along with the icon_size / icon_color options. The {logo} token now links a bundled src/assets/images/claude.png via an  tag, and users can of course drop in any Nerd Font glyph or their own  instead.

4. Fair point. I studied how the existing widgets/services actually work this time (used server_monitor for the service and github for the menu) instead of trusting the generated code, and reworked it from there. Hopefully this pass lines up much better with how YASB is structured happy to adjust anything that still looks off.

5. menu instead of card renamed throughout: the config block is menu:, the callback is toggle_menu, and the popup class is .claude-usage-menu, mirroring the github widget.

PR diff is down to 6 files with no schema.json. Let me know if you'd like any further changes

amnweb:
I still don't get why you need a hardcoded image there, or why all this mess with it and the `{logo}` placeholder?
You can use an image anyway in the label `label: " {five_hour}"`
`CallbacksConfig` is also wrong, you could use someting like 
```py
class ClaudeUsageCallbacksConfig(CallbacksConfig):
    on_left="toggle_menu",
    on_middle="do_nothing",
    on_right="toggle_label",
```
and than
```py
class ClaudeUsageConfig(CustomBaseModel):
    ...
    callbacks: ClaudeUsageCallbacksConfig = ClaudeUsageCallbacksConfig()
    ```

**Comment by ManaphatDev** [comment]
Yeah you're right, that was overcomplicated 

Cleaned it up:
Dropped the bundled image and the whole {logo} placeholder thing. The icon is just part of the label now — a Nerd Font glyph by default, and if someone wants the actual Claude logo they can just drop in  like you said. Added a note in the docs pointing to svgl.app for the logo.
Fixed the callbacks too moved them into a ClaudeUsageCallbacksConfig(CallbacksConfig) subclass instead of passing kwargs, so the generated schema picks up the right defaults.
Pushed. Thanks for bearing with me on this

**Comment by amnweb** [comment]
Thanks, a few more suggestions:

`QProgressBar` is not a good solution for showing progress bars because there is really annoying calculation involved, and if you have a border radius for the progress bar and the value is low, you end up with a square fill. That is a known issue in Qt. Check the Copilot widget, where the progress bar is drawn manually.

Another suggestion: clean up all validation comments. We really do not need them there.

Also, I'm still not sure about `CLAUDE_ICON = "\U000f06a9"`. In this way, you force users to install a specific font to display these icons. If you "must" show this icon, keep it configurable. If you don't need it, remove it. I don't know if any popup shows icons in this way in the project, to be honest.

**Comment by ManaphatDev** [comment]
Thanks a ton for the review — super helpful, and all three make sense.

**1. Progress bar** — yeah, I get the rounded-corner square-fill thing. I'll ditch `QProgressBar` and go with the track + child `.fill` QFrame setup like the Copilot widget (including the `max(fill_width, height)` trick so it stays rounded at low values). Bonus: it ends up fully CSS-styleable.

**2. Validation comments** — yep, stripping those out.

**3. Icon** — fair point, I don't want to force a font on anyone either. I've actually already moved it off the Nerd Font glyph and render it as an SVG with `QtSvg` (same as `launchpad` / `open_meteo` / `quick_launch`), so no font needed and it's the real logo instead of a code-point. Still configurable (color/size) and totally optional — and if you'd rather it ship without an icon by default, just say the word and I'll pull it.

I'll push the updated version shortly. Cheers!

**Comment by amnweb** [comment]
I really don’t think you need icons for this. Make it a header like we have in all other popups, like Copilot, GitHub, etc. Icons in Open-Meteo or Quick Launch are used for a different purpose, to be honest.

**Comment by ManaphatDev** [comment]
Makes sense dropping the icon and going with a plain header like the other popups. Pushing shortly

**Comment by ManaphatDev** [comment]
Pushed an update addressing all three points:

- Draw the menu progress bar as a track + child `.fill` QFrame instead of `QProgressBar`, keeping the fill at least as wide as it is tall so the rounded corners survive at low values (the `QProgressBar::chunk` square-fill issue).
- Plain text menu header like the other popups (Copilot, GitHub) — dropped the icon.
- Removed the explanatory comments from the validation schema.

image
image

**Comment by amnweb** [comment]
OK, thanks. Just please remove lines like this one:

`# -- popup menu -------------------------------------------------------`

I really hate this type of comment when AI adds it. And please clean up the commit history, all of this can be one commit. When you think it is ready to merge, please ping me.

**Comment by ManaphatDev** [comment]
Done 

Removed all those banner comments 
Squashed the whole branch down to a single commit.
I think it's good to merge now  ping! Thanks for taking the time to review this.

**Comment by amnweb** [comment]
This AI which you use is terrible. Once you fix all of these errors, and there are probably more, tell me to check again. I really don't want to waste time checking and testing this PR anymore until everything is fixed.
```
  File "E:\Project\css\src\core\widgets\services\claude_usage\claude_api.py", line 67, in fetch_usage
    with open(cred_path, encoding="utf-8") as f:
         ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\amn\\.claude\\.credentials.json'
```

ManaphatDev:
Thanks for flagging it, and sorry for the noise.

You're right that it was printing that traceback. Cause: when
`~/.claude/.credentials.json` doesn't exist (Claude Code not signed in), the fetch
logged the exception with `exc_info=True`. The call was already wrapped in try/except
so the widget degraded to `--` instead of crashing the bar — but dumping a full
traceback on every poll for an expected "not signed in" state is wrong and reads
exactly like a crash.

Fixed it to log like `server_monitor` does for expected failures — one concise debug
line, no traceback (same change applied to the cache-write fallback):

```python
except Exception as e:
    logger.debug("usage fetch failed: %s", e)

amnweb:
OK, another suggestion: if we keep the hardcoded character ↻, I'm sure someone will open an issue asking how to change it 😞
Maybe we can remove this icon from the label?
If you don't plan to change this PR, let me know so I can merge it and release a new version soon.

ManaphatDev:
Sorry about the noisy traceback in the last round — that was an over-eager exc_info log on an expected "not signed in" path, now fixed so it logs quietly at debug level.

On the icon: good point — removed the hardcoded ↻ from the reset label so there's nothing baked in that people would need to ask about. The bar label was already plain text (Claude {five_hour}%) with no icon, so users can add their own glyph via a <span> in label if they want one.

Pushed both changes. Ready to merge whenever you are — thanks for the patience and the review!  🙇‍♂️

PR #957, opened 2026-06-09
fix(glazevm): web socket connection on windows lock/sleep

joelastic:
## Summary

  - Add WebSocket ping/keepalive to GlazeWM client to detect and recover from stale connections after screen lock, sleep, or network interruption

  ## Problem

  After locking the screen for a while and coming back, YASB hangs — GlazeWM workspace widgets stop updating.

  **Root cause:** The WebSocket connection to GlazeWM silently dies during screen lock/sleep. TCP half-open state means QWebSocket never fires `disconnected` or `errorOccurred`, so the socket still reports
  `ConnectedState`. The reconnect timer never triggers, and the widget stays frozen with stale data.

  ## Fix

  Added a ping/pong keepalive mechanism to `GlazewmClient`:

  - A ping timer (default 30s) runs while connected, sending WebSocket ping frames
  - If a pong is not received before the next ping tick, the connection is considered stale
  - `abort()` is used instead of `close()` to immediately drop the dead socket (close would attempt a handshake on a dead connection and hang)
  - The reconnect timer then kicks in and re-establishes the connection with fresh event subscriptions

  **Worst-case recovery time:** ~34s after unlock (one missed ping cycle + reconnect interval).

  ## Tested

  - [x] Start YASB with GlazeWM running — verify workspaces widget works normally
  - [x] Lock screen, wait 2+ minutes, unlock — verify workspaces recover and update within ~30s
  - [x] Stop GlazeWM while YASB is running — verify reconnect works when GlazeWM restarts
  - [x] Check logs for `No pong received, connection stale. Forcing reconnect.` after lock/unlock recovery

### DIFF
--- a/src/core/widgets/base.py
+++ b/src/core/widgets/base.py
@@ -152,30 +152,37 @@ class BaseWidget(QWidget):
         self,
         content: str,
         content_alt: str | None = None,
+        label_placeholder: str | None = None,
+        hide_icons: bool = False,
     ):
         def process_content(content: str, is_alt: bool = False) -> list[QLabel]:
-            label_parts = re.split("(<span.*?>.*?</span>)", content)
+            label_parts = re.split(r"(<span.*?>.*?</span>)", content)
             label_parts = [part for part in label_parts if part]
             widgets: list[QLabel] = []
             for part in label_parts:
                 part = part.strip()
                 if not part:
                     continue
-                if "<span" in part and "</span>" in part:
+                is_icon = "<span" in part and "</span>" in part
+                if is_icon:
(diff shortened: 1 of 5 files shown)


amnweb:
Hi, 
Can you clean up this commits history? It can be a single commit.

joelthegerman:
> Hi, 
> 
> Can you clean up this commits history? It can be a single commit.
> 
> 

Hi,

I will fix it as soon as possible. Sorry for the inconvenience.

joelthegerman:
Automatically closed because of some merge conflicts i got and i had to reset branch, created a new one here https://github.com/amnweb/yasb/pull/961

@amnweb

PR #967, opened 2026-06-12
feat(claude_usage): refresh button, expired-token indicator & configurable reset display

jondmarien:
## Summary

The Claude Usage widget currently has no way to force a re-fetch — when the data goes stale (e.g. the OAuth token in `.credentials.json` expired for a while and every poll silently fell back to the disk cache), the widget keeps showing old percentages with no recourse besides restarting the bar.

This PR adds:

- **`refresh` widget callback** — bindable to any mouse button (e.g. `on_middle: "refresh"`), forces an immediate fetch that bypasses the cache TTL via a new `ClaudeUsageService.refresh_now()`.
- **Refresh button in the popup menu header** (nerd-font `\U000f0450` glyph, `ClickableLabel` following the pattern used in `wifi_widgets.py`), with a "Refresh now" tooltip.
- **Live menu updates** — the popup sections now redraw in place when fresh data arrives while the menu is open, instead of showing stale values until reopened.

In-flight fetches are still deduplicated (`refresh_now()` is a no-op while a worker is running), so the button can't stack requests.

## Styling notes

The menu header changed from a single `QLabel.header` to a `QFrame.header` row containing `QLabel.header-title` + `ClickableLabel.refresh`. Since Qt stylesheets don't propagate font/color to children, existing `.claude-usage-menu .header` font rules should be duplicated onto `.header-title`. Example CSS:

```css
.claude-usage-menu .header .header-title {
    font-size: 14px;
    font-weight: 600;
    color: #cdd6f4;
}
.claude-usage-menu .header .refresh {
    font-size: 15px;
    color: rgba(205, 214, 244, 0.55);
    padding: 0 2px;
}
.claude-usage-menu .header .refresh:hover {
    color: #fab387;
}
```

Happy to update the wiki page ((Widget)-Claude-Usage) to document the new callback and CSS classes if this is accepted.

## Testing

- Built the modified modules into a local v2.0.3 install (Python 3.14, frozen `library.zip`) and verified: middle-click refresh, menu button refresh, in-place menu redraw on data arrival, and no regressions on toggle_label/toggle_menu.
- `ruff check` / `ruff format` clean (no new findings vs. main).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/core/widgets/services/claude_usage/claude_api.py
+++ b/src/core/widgets/services/claude_usage/claude_api.py
@@ -23,11 +23,30 @@ EMPTY_RECORD: dict[str, Any] = {
     "seven_raw": None,
     "seven_reset_iso": None,
     "fetched_at": 0,
+    "token_expired": False,
 }
 
 
 def _claude_config_dir() -> str:
     return os.environ.get("CLAUDE_CONFIG_DIR") or os.path.join(os.path.expanduser("~"), ".claude")
+
+
+def _token_expired() -> bool:
+    """True when Claude Code's OAuth access token has expired (so a refresh is needed).
+
+    Reads only the ``expiresAt`` timestamp (ms epoch); the token itself is never
+    touched here. Returns False when the value is missing/unreadable so a parsing
+    quirk never produces a false 'expired' warning.
+    """
+    try:
+        cred_path = os.path.join(_claude_config_dir(), ".credentials.json")
+        with open(cred_path, encoding="utf-8") as f:
+            expires_at = json.load(f)["claudeAiOauth"].get("expiresAt")
+        if not expires_at:
+            return False
+        return (expires_at / 1000) < time.time()
+    except Exception:
+        return False
 
 
 def _cache_path() -> str:
@@ -57,9 +76,14 @@ def fetch_usage(cache_path: str, cache_ttl: int) -> dict[str, Any]:
     returned so the widget keeps showing the most recent known values. The OAuth
     token is read from Claude Code's credentials store and is never logged.
     """
+    # Recomputed live on every call (independent of the usage cache) so the
+    # expired-token warning is accurate even when serving a stale record.
+    token_expired = _token_expired()
+
     cache = _read_cache(cache_path)
     now = int(time.time())
     if cache and (now - int(cache.get("fetched_at", 0))) < cache_ttl:
+        cache["token_expired"] = token_expired
         return cache
 
     try:
@@ -84,14 +108,19 @@ def fetch_usage(cache_path: str, cache_ttl: int) -> dict[str, Any]:
             "seven_raw": seven_raw,
             "seven_reset_iso": payload["seven_day"].get("resets_at"),
             "fetched_at": now,
+            # A successful fetch proves the token is currently valid.
+            "token_expired": False,
         }
         _write_cache(cache_path, record)
         return record
     except Exception as e:
         logger.debug("usage fetch failed: %s", e)
         if cache:
+            cache["token_expired"] = token_expired
             return cache
-        return dict(EMPTY_RECORD)
+        record = dict(EMPTY_RECORD)
+        record["token_expired"] = token_expired
+        return record
 
 
 class _UsageWorker(QThread):
--- a/src/core/widgets/services/claude_usage/claude_api.py
+++ b/src/core/widgets/services/claude_usage/claude_api.py
@@ -165,9 +165,16 @@ class ClaudeUsageService(QObject):
             self.deleteLater()
 
     def _tick(self) -> None:
+        self._start_worker(self._cache_ttl)
+
+    def refresh_now(self) -> None:
+        """Force an immediate fetch, bypassing the cache TTL."""
+        self._start_worker(0)
+
+    def _start_worker(self, cache_ttl: int) -> None:
         if self._worker is not None:
             return  # a fetch is already in flight
-        worker = _UsageWorker(self._cache_path, self._cache_ttl, self)
+        worker = _UsageWorker(self._cache_path, cache_ttl, self)
         worker.data_ready.connect(self._on_data)
         worker.finished.connect(self._on_finished)
         self._worker = worker
--- a/src/core/widgets/yasb/claude_usage.py
+++ b/src/core/widgets/yasb/claude_usage.py
@@ -257,6 +257,7 @@ class ClaudeUsageWidget(BaseWidget):
             layout = self._menu_layout
             for frame in getattr(self, "_section_frames", []):
                 layout.removeWidget(frame)
+                frame.hide()
                 frame.deleteLater()
             self._add_menu_sections(layout)
             menu.adjustSize()
@@ -279,13 +280,13 @@ class ClaudeUsageWidget(BaseWidget):
         self._menu_layout = layout
 
         header = QFrame()
-        header.setProperty("class", "header")
+        header.setProperty("class", "header-row")
         header_layout = QHBoxLayout(header)
         header_layout.setContentsMargins(0, 0, 0, 0)
         header_layout.setSpacing(0)
 
         header_title = QLabel("Claude Usage")
-        header_title.setProperty("class", "header-title")
+        header_title.setProperty("class", "header")
         header_layout.addWidget(header_title)
         header_layout.addStretch()
 
--- a/src/core/widgets/yasb/claude_usage.py
+++ b/src/core/widgets/yasb/claude_usage.py
@@ -10,6 +10,9 @@ from core.validation.widgets.yasb.claude_usage import ClaudeUsageConfig
 from core.widgets.base import BaseWidget
 from core.widgets.services.claude_usage.claude_api import ClaudeUsageService
 
+# Usage-level CSS classes (shared by the popup bars and the optional bar-percent colouring).
+_LEVEL_CLASSES = frozenset({"low", "medium", "high", "critical", "unknown"})
+
 
 class UsageBar(QFrame):
     """A rounded progress bar drawn as a track QFrame with a child `.fill` QFrame.
@@ -190,13 +193,22 @@ class ClaudeUsageWidget(BaseWidget):
 
     @staticmethod
     def _level_class(value: Any) -> str:
+        """Usage band: low < 50 ≤ medium < 65 ≤ high < 80 ≤ critical (green/yellow/orange/red)."""
         if not isinstance(value, (int, float)):
             return "unknown"
         if value >= 80:
+            return "critical"
+        if value >= 65:
             return "high"
         if value >= 50:
             return "medium"
         return "low"
+
+    @staticmethod
+    def _apply_level_class(widget: QLabel, level: str) -> None:
+        """Set ``widget``'s usage level (low/medium/high) while keeping its base class."""
+        base = " ".join(t for t in (widget.property("class") or "label").split() if t not in _LEVEL_CLASSES)
+        widget.setProperty("class", f"{base} {level}")
 
     def _toggle_label(self) -> None:
         self._show_alt_label = not self._show_alt_label
@@ -230,6 +242,11 @@ class ClaudeUsageWidget(BaseWidget):
             # Hide empty labels so an unused placeholder (e.g. {stale} when the token is
             # valid) doesn't leave a constant gap from its margin/spacing.
             current_widget.setVisible(bool(rendered))
+            # Optionally colour the percentage by usage level (same thresholds as the bars).
+            if self.config.colorize_percent:
+                window = "five" if "{five_hour}" in part else "seven" if "{seven_day}" in part else None
+                if window is not None:
+                    self._apply_level_class(current_widget, self._level_class(self._data.get(window)))
             if self.config.tooltip:
                 tip = f"Claude usage — 5h: {values['five_hour']}% · 7d: {values['seven_day']}%"
                 if self._data.get("token_expired"):
--- a/src/core/widgets/yasb/claude_usage.py
+++ b/src/core/widgets/yasb/claude_usage.py
@@ -151,9 +151,9 @@ class ClaudeUsageWidget(BaseWidget):
 
     @staticmethod
     def _fmt_weekday(iso: str | None, with_date: bool = False) -> str:
-        """Absolute reset as a local weekday + time, e.g. 'Sat 6:00 AM'; '--' when unknown.
+        """Absolute reset as a local weekday + time, e.g. 'Sat @ 6:00 AM'; '--' when unknown.
 
-        With ``with_date`` the month/day is included ('Sat, Jun 13 6:00 AM') so two windows
+        With ``with_date`` the month/day is included ('Sat, Jun 13 @ 6:00 AM') so two windows
         resetting on the same weekday can be told apart.
         """
         if not iso:
@@ -163,7 +163,7 @@ class ClaudeUsageWidget(BaseWidget):
             hour12 = local.hour % 12 or 12
             ampm = "AM" if local.hour < 12 else "PM"
             day = f"{local:%a, %b} {local.day}" if with_date else f"{local:%a}"
-            return f"{day} {hour12}:{local.minute:02d} {ampm}"
+            return f"{day} @ {hour12}:{local.minute:02d} {ampm}"
         except Exception:
             return "--"
 
--- a/src/core/widgets/yasb/claude_usage.py
+++ b/src/core/widgets/yasb/claude_usage.py
@@ -2,8 +2,7 @@ import re
 from datetime import UTC, datetime
 from typing import Any
 
-from PyQt6.QtCore import Qt, pyqtSignal
-from PyQt6.QtWidgets import QFrame, QHBoxLayout, QLabel, QVBoxLayout
+from PyQt6.QtWidgets import QFrame, QHBoxLayout, QLabel, QPushButton, QVBoxLayout
 
 from core.utils.tooltip import set_tooltip
 from core.utils.utilities import PopupWidget, refresh_widget_style
@@ -35,19 +34,6 @@ class UsageBar(QFrame):
     def resizeEvent(self, event):
         super().resizeEvent(event)
         self._update_fill()
-
-
-class ClickableLabel(QLabel):
-    clicked = pyqtSignal()
-
-    def __init__(self, text: str, parent: QFrame | None = None):
-        super().__init__(text, parent)
-        self.setCursor(Qt.CursorShape.PointingHandCursor)
-
-    def mousePressEvent(self, event):
-        if event.button() == Qt.MouseButton.LeftButton:
-            self.clicked.emit()
-        super().mousePressEvent(event)
 
 
 class ClaudeUsageWidget(BaseWidget):
@@ -280,21 +266,21 @@ class ClaudeUsageWidget(BaseWidget):
         self._menu_layout = layout
 
         header = QFrame()
-        header.setProperty("class", "header-row")
+        header.setProperty("class", "header")
         header_layout = QHBoxLayout(header)
         header_layout.setContentsMargins(0, 0, 0, 0)
         header_layout.setSpacing(0)
 
-        header_title = QLabel("Claude Usage")
-        header_title.setProperty("class", "header")
-        header_layout.addWidget(header_title)
+        title_label = QLabel("Claude Usage")
+        title_label.setProperty("class", "text")
+        header_layout.addWidget(title_label)
         header_layout.addStretch()
 
-        refresh_button = ClickableLabel("\U000f0450")
-        refresh_button.setProperty("class", "refresh")
-        set_tooltip(refresh_button, "Refresh now")
-        refresh_button.clicked.connect(self._refresh)
-        header_layout.addWidget(refresh_button)
+        refresh_btn = QPushButton("\U000f0450")
+        refresh_btn.setProperty("class", "refresh")
+        set_tooltip(refresh_btn, "Refresh now")
+        refresh_btn.clicked.connect(self._refresh)
+        header_layout.addWidget(refresh_btn)
 
         layout.addWidget(header)
         self._add_menu_sections(layout)
--- a/src/core/widgets/yasb/claude_usage.py
+++ b/src/core/widgets/yasb/claude_usage.py
@@ -85,12 +85,16 @@ class ClaudeUsageWidget(BaseWidget):
     def _refresh(self) -> None:
         self._service.refresh_now()
 
+    # Warning glyph (nf-fa-warning) shown via {stale} when the OAuth token has expired.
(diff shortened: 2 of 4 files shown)


ManaphatDev:
Hey @jondmarien! Really appreciate this — the stale-cache issue was something I'd noticed but hadn't gotten around to. Applied the CSS locally and it looks great. The live menu redraw is a nice touch too.

src/core/widgets/yasb/claude_usage.py
```diff
@@ -230,21 +276,27 @@ def _build_menu(self) -> None:
         layout = QVBoxLayout(self._menu)
         layout.setContentsMargins(0, 0, 0, 0)
         layout.setSpacing(0)
+        self._menu_layout = layout
 
-        header = QLabel("Claude Usage")
+        header = QFrame()
         header.setProperty("class", "header")
```
ManaphatDev, line 350: Moving `class="header"` from the title `QLabel` to this new `QFrame` container silently breaks existing stylesheets: anyone on v2.0.3 who styled `.claude-usage-menu .header` (font-size/color/weight) loses their header styling, since QSS doesn't propagate font/color from a frame to its children.

Could you keep `class="header"` on the title label and give the container row a new class instead (e.g. `header-row`)? Then existing configs keep working and only the new `.refresh` class needs documenting.

src/core/widgets/yasb/claude_usage.py
```diff
@@ -217,6 +236,33 @@ def _build_section(self, title: str, value: Any, raw: Any, reset_iso: str | None
...
+    def _refresh_menu_sections(self) -> None:
+        """Redraw the popup sections in place when fresh data arrives while it is open."""
+        menu = self._menu
+        try:
+            if menu is None or not menu.isVisible():
+                return
+            layout = self._menu_layout
+            for frame in getattr(self, "_section_frames", []):
+                layout.removeWidget(frame)
+                frame.deleteLater()
```
ManaphatDev, line 328: Nit: `removeWidget()` leaves the frame parented and visible until `deleteLater()` is processed on the next event-loop pass, so old and new sections can briefly coexist. A `frame.hide()` before `deleteLater()` rules out any one-frame flicker during the in-place redraw.

src/core/widgets/services/claude_usage/claude_api.py
```diff
@@ -165,9 +165,16 @@ def release(self) -> None:
...
     def _tick(self) -> None:
+        self._start_worker(self._cache_ttl)
+
+    def refresh_now(self) -> None:
+        """Force an immediate fetch, bypassing the cache TTL."""
+        self._start_worker(0)
+
+    def _start_worker(self, cache_ttl: int) -> None:
         if self._worker is not None:
             return  # a fetch is already in flight
```
ManaphatDev, line 205: Noting for the record (fine to leave as-is): clicking refresh while a timer-tick fetch is already in flight is a silent no-op — and that in-flight worker may be serving from cache, so the click can produce no fresh fetch with no feedback. The window is tiny; just flagging it in case anyone ever reports "refresh didn't do anything".

ManaphatDev (commented):
Thanks for this — refresh was a known gap (the widget intentionally degrades to the cached value when the token is stale, so a manual re-fetch is a nice escape hatch). Went through both files; the service change is clean: `refresh_now()` → `_start_worker(0)` is the minimal correct way to bypass the TTL while keeping the error-fallback-to-cache path and in-flight dedup intact, and a manual refresh rewrites the cache so the next timer tick serves it without an extra network call. The `RuntimeError` handling in `_refresh_menu_sections` also correctly covers `PopupWidget` deleting itself on close.

One real request inline (the `.header` class change breaks existing stylesheets), plus a couple of take-or-leave nits. And yes please on the wiki update — the new callback and the header CSS classes should be documented before this lands.

jondmarien:
Thanks for the review @ManaphatDev! Both points addressed in 745e91c:

- `class="header"` is back on the title `QLabel`, so existing `.claude-usage-menu .header` rules keep working untouched. The container row got the new `header-row` class as suggested — the only new classes to document are `.header-row` and `.refresh`.
- Added `frame.hide()` before `deleteLater()` in the in-place redraw.

On the refresh-during-in-flight-fetch note: agreed it's a tiny window — leaving as-is per your comment, but happy to add a queued-refresh flag if it ever comes up in reports.

**Wiki:** the wiki repo only accepts pushes from collaborators, so here's the proposed update to `(Widget)-Claude-Usage` as a diff — covers the `refresh` callback, the menu refresh button, the new CSS classes (style reference + example style), and a note that `refresh` can't renew an expired OAuth token (only running Claude Code does):

<details>
<summary>Proposed wiki diff</summary>

```diff
diff --git a/widgets/(Widget)-Claude-Usage.md b/widgets/(Widget)-Claude-Usage.md
index 5a0c081..6c6a3e5 100644
--- a/widgets/(Widget)-Claude-Usage.md
+++ b/widgets/(Widget)-Claude-Usage.md
@@ -40,7 +40,7 @@ claude_usage:
     cache_ttl: 120
     callbacks:
       on_left: "toggle_menu"    # open the usage menu
-      on_middle: "do_nothing"
+      on_middle: "refresh"      # force an immediate re-fetch, bypassing the cache
       on_right: "toggle_label"  # switch the bar text between 5h and 7d
     menu:
       blur: true
@@ -60,7 +60,7 @@ claude_usage:
 - **update_interval:** How often the bar label and reset countdown are refreshed, in seconds (30–3600).
 - **cache_ttl:** How long a fetched result is cached on disk before the usage endpoint is queried again. Because the endpoint is rate-limited (HTTP 429), the widget serves the last cached value on any error instead of going blank.
 - **tooltip:** Whether to show a summary tooltip on hover.
-- **callbacks:** Mouse-click callbacks. Built-in actions: `toggle_menu` (open/close the popup menu), `toggle_label` (swap between `label` and `label_alt`), `do_nothing`, and `exec`.
+- **callbacks:** Mouse-click callbacks. Built-in actions: `toggle_menu` (open/close the popup menu), `toggle_label` (swap between `label` and `label_alt`), `refresh` (force an immediate re-fetch of the usage data, bypassing `cache_ttl`), `do_nothing`, and `exec`.
 - **menu:** A dictionary specifying the popup menu settings:
   - **blur:** Enable blur effect for the menu.
   - **round_corners:** Enable round corners (not supported on Windows 10).
@@ -77,6 +77,20 @@ The widget reuses Claude Code's existing OAuth session. It reads the access toke
 that environment variable is set) and never logs or stores it elsewhere. If you are not
 signed in to Claude Code, the widget shows `--` until you sign in.
 
+> [!NOTE]
+> Only Claude Code itself renews the OAuth token. If the token has expired (e.g. you have
+> not used Claude Code in a while), the usage endpoint rejects the request and the widget
+> keeps serving the last cached values until the token is refreshed — running any Claude
+> Code command renews it. The `refresh` callback / menu button forces a re-fetch but cannot
+> renew an expired token.
+
+## Refresh
+
+The popup menu header contains a refresh button (`\U000f0450`) that forces an immediate
+re-fetch, bypassing `cache_ttl`. The same action is available as the `refresh` callback for
+any mouse button. While the menu is open, its sections redraw in place whenever fresh data
+arrives. If a scheduled fetch is already in flight, a refresh request is ignored.
+
 ## Widget Style
 ```css
 .claude-usage {}
@@ -85,7 +99,10 @@ signed in to Claude Code, the widget shows `--` until you sign in.
 .claude-usage .label {}
 /* Popup menu */
 .claude-usage-menu {}
+.claude-usage-menu .header-row {}    /* header container (title + refresh button) */
 .claude-usage-menu .header {}        /* "Claude Usage" title */
+.claude-usage-menu .refresh {}       /* refresh button in the header */
+.claude-usage-menu .refresh:hover {}
 .claude-usage-menu .section {}
 .claude-usage-menu .section .title {}
 .claude-usage-menu .section .progress {}               /* progress-bar track */
@@ -116,11 +133,21 @@ signed in to Claude Code, the widget shows `--` until you sign in.
     background-color: #1e1e2e;
     min-width: 260px;
 }
+.claude-usage-menu .header-row {
+    padding: 14px 16px 10px 16px;
+}
 .claude-usage-menu .header {
     color: #cdd6f4;
     font-size: 15px;
     font-weight: bold;
-    padding: 14px 16px 10px 16px;
+}
+.claude-usage-menu .refresh {
+    color: #6c7086;
+    font-size: 15px;
+    padding: 0 2px;
+}
+.claude-usage-menu .refresh:hover {
+    color: #fab387;
 }
 .claude-usage-menu .section {
     padding: 4px 16px 12px 16px;
```

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jondmarien:
On a personal note -- I use YASB daily and love building custom things for it. One of my favorites is a Burp Suite indicator that shows on the bar when Burp is running (currently a CustomWidget + PowerShell script combo). Once I've finished testing it I'd like to clean it up and submit it as a proper widget PR too. Happy to keep contributing!

amnweb:
I think your original idea for the layout was right:
```
header
  ->title
  -> button

```
Also, for the refresh button, you can use a `QPushButton` instead of a clickable label. Check how I built the popup for memory and CPU
https://github.com/amnweb/yasb/blob/9b047ec6edaa59ad60a7d91c14b33d77b8095b62/src/core/utils/stat_popup.py#L229-L244
I think that is the right approach for the header. If you think it is not, please tell me.

About the wiki, it is updated automatically and generated from the docs folder when a stable release is published. So you should edit the files in the docs folder, and once a stable release is published, the wiki will be generated.

jondmarien:
> I think your original idea for the layout was right:
> 
> ```
> header
>   ->title
>   -> button
> ```
> 
> Also, for the refresh button, you can use a `QPushButton` instead of a clickable label. Check how I built the popup for memory and CPU
> 
> https://github.com/amnweb/yasb/blob/9b047ec6edaa59ad60a7d91c14b33d77b8095b62/src/core/utils/stat_popup.py#L229-L244
> 
> I think that is the right approach for the header. If you think it is not, please tell me.
> 
> About the wiki, it is updated automatically and generated from the docs folder when a stable release is published. So you should edit the files in the docs folder, and once a stable release is published, the wiki will be generated.

Hi @amnweb, yes I believe that the original idea for the layout would be better. I will revert the change, and edit the files in the docs folder, so they get automatically generated, like you mentioned. Thanks for the comment!

src/core/widgets/yasb/claude_usage.py
jondmarien: Reworked the header to follow the stat_popup.py convention per @amnweb: container is QFrame.header, the title is a QLabel.text, and the refresh button is now a QPushButton.refresh (dropped the custom ClickableLabel). New classes documented in docs/. (03f5725)

src/core/widgets/yasb/claude_usage.py
jondmarien: Done — added frame.hide() before deleteLater() in the in-place redraw (745e91c) so old and new sections can't briefly coexist.

jondmarien:
Pushed in 03f5725. The header now follows the `stat_popup.py` pattern from the memory/CPU popups: a `QFrame` with class `header` containing a `QLabel.text` title and a `QPushButton.refresh` button (the custom `ClickableLabel` is gone). 

And thanks for clarifying the wiki flow — I moved the docs changes (refresh callback, the new menu button, and the `.header .text` / `.header .refresh` CSS classes) into `docs/widgets/(Widget)-Claude-Usage.md` so they'll generate on the next stable release. The standalone wiki-diff comment is now obsolete.

jondmarien:
I'm also noticing in testing that if the token is invalidated for over a long period of time (~8hrs), the `refresh` button does nothing. I am looking into this for a solution, and will update the PR when i find one.

jondmarien:
I've found some restrictions:

The widget can't self-heal the invalidated token. I looked at having it refresh the token itself, but the OAuth token endpoint sits behind Cloudflare (403/1010) and scripting around that is credential misuse, and as the harness correctly blocked me from trying, I'm not going to pursue it.

Only a fresh claude process refreshes the disk token. So the realistic fix is to run a trivial `claude -p` on a schedule.

Not sure how we can tie this into the widget. If you have any ideas, @amnweb, I'd be interested in hearing them!!!

jondmarien:
Added a small expired-token indicator (a866150). When Claude Code's OAuth token expires, the usage endpoint 429s and the widget keeps showing the last cached values with no signal that they're stale -- so I added a live `token_expired` flag (computed from the credentials' `expiresAt` on each fetch, independent of the usage cache) and a `{stale}` label placeholder that renders a warning glyph (styleable via `.claude-usage .stale`) only while the token is expired, empty otherwise. The tooltip also hints to run a `claude` command to refresh. It's purely a cue -- only Claude Code can renew the token. Docs updated in `docs/` accordingly.

This is what I have come up with, let me know if you think this is okay to keep in.

ManaphatDev:
Tested this locally against my real setup (Python 3.14, run from source on top of v2.0.3) — looks good to merge from my side.

What I verified:

- **`refresh` callback** — bound it to `on_middle` and it fires cleanly now. Worth noting the contrast: on stock v2.0.3 a `refresh` binding throws `KeyError: 'refresh'` in `_run_callback` every time, since the callback didn't exist; with this PR it's registered and works.
- **Menu refresh button** — renders correctly in the header (`QFrame.header` → `QLabel.text` + `QPushButton.refresh`, the `\U000f0450` glyph), matching the `stat_popup.py` convention you asked for:

  claude usage menu with refresh button

- **`refresh_now()`** — `_start_worker(0)` bypasses the TTL and forces a real fetch, while the `self._worker is not None` guard keeps in-flight dedup intact, so the button can't stack requests.
- **`token_expired`** — recomputed live from `expiresAt` on each `fetch_usage` (independent of the usage cache), and returns `False` on missing/unreadable so it never false-positives. The `{stale}` placeholder renders the warning glyph only while expired, empty otherwise.
- **In-place menu redraw** — `_refresh_menu_sections()` is a no-op while the popup is hidden and redraws without exceptions while open; the `RuntimeError` guard correctly covers the popup being destroyed on close.
- `ruff check` / `ruff format` clean.

Two optional nits, neither blocking:

1. The `{stale}` span label isn't hidden when empty — with the example CSS's `.stale { margin-left: 4px }`, a valid (non-expired) token still leaves a small constant gap, since `_update_label` calls `setText("")` rather than hiding the label. Could `hide()` it when the formatted value is empty, or drop the margin. Only affects setups that copy the example `{stale}` label.
2. Trivial: `STALE_ICON` is declared mid-class between `_refresh` and `_format_values` — reads a bit more naturally up top with the other class attributes.

Nice work — the live `token_expired` indicator is a good touch. 👍

jondmarien:
I did notice the constant gap when the Stale label doesn't show, you can see it very slightly.
image
I'll fix this tonight or tomorrow.

Also going to fix the wording for the 7-day reset "Resets in Day Time". It is incorrect grammar.

jondmarien:
Pushed both fixes (67b596d):

**1. Empty `{stale}` gap (your nit #1).** `_update_label` now hides a label when its formatted text is empty, so a valid (non-expired) token no longer leaves the constant ~4px gap from the example CSS's `.stale { margin-left }`. When the token expires the label is shown again. Also moved `STALE_ICON` up with the other class attributes (nit #2).

**2. Menu reset wording.** The 7-day footer read `Resets in Sat 6:00 AM`, which is grammatically off (the relative "in" + an absolute weekday/time). Added a `reset_format` option:
- `absolute` (default) → `Resets on Sat 6:00 AM`
- `relative` → `Resets in 6d 21h`

The exact reset timestamp is still shown on the line below either way, and the bar-label `{five_hour_reset}`/`{seven_day_reset}` placeholders are unchanged (kept `_fmt_reset` as-is). Docs updated in `docs/`.

`ruff check`/`format` clean (the 4 `UP038` `isinstance` warnings are pre-existing on the file, untouched by this diff).

jondmarien:
Much better, imo
also, added the five_hour_reset indicator, thank you for that haha (from your setup, @ManaphatDev)

image

jondmarien:
I just noticed another edge-case where where the 7 day reset timer and 5 day reset timer are on the same day (or rather looks like it, 1 week apart). Going to add a date indicator/label. Optional, but enabled by default. What do you guys think?

jondmarien:
Small follow-up (78712b0): in `absolute` mode the 5-hour and 7-day windows can both read `Resets on Sat …` on *different* Saturdays, which is ambiguous. Added `reset_show_date` (default `true`) to include the month/day -- `Resets on Sat, Jun 13 6:00 AM` -- so the two windows are distinguishable at a glance. No effect in `relative` mode. Docs updated.

jondmarien:
Ah, much better

image

jondmarien:
Accidentally closed with comment, apologies. Adding a tiny commit to add a `@` between Date + Time, so "Resets in Date @ Time"

jondmarien:
Tiny polish (d133967): put an `@` between the date and time in the absolute reset line — `Resets on Sat, Jun 13 @ 6:00 AM` — reads more clearly than running them together.

jondmarien:
If you like, I can squash the commits into one, I noticed the main author was mentioning this in some PRs. Let me know if you would like this to happen.

jondmarien:
Made the reset format **per-window** (3bc5674): replaced the single `reset_format` with `five_hour_reset_format` and `seven_day_reset_format`. The defaults now fit each timescale — the near-term 5-hour window as a countdown (`Resets in 4h 11m`) and the multi-day 7-day window as a date (`Resets on Sat, Jun 20 @ 2:59 AM`). `reset_show_date` still applies to whichever window uses `absolute`. Docs updated.

image

Figure 1. Showing two different reset window timers (relative for 5hr by default, absolute for 7hr by default)

ManaphatDev:
> Ah, much better
> 
> image

Soo cool

jondmarien:
> > Ah, much better
> > image
> 
> Soo cool

Thank you for making this widget!

ManaphatDev:
> > > Ah, much better
> > > image
> > 
> > 
> > Soo cool
> 
> Thank you for making this widget!

I built this because I primarily use Claude Code (CLI), and VS Code already has usage info in the extension — I just wanted the same visibility on my bar YASB  while coding. Glad it's useful for others too 😄

Thanks for building on top of it, the refresh button was exactly what was missing. Looking forward to getting this merged!

jondmarien:
> > > > Ah, much better
> > > > image
> > > 
> > > 
> > > Soo cool
> > 
> > 
> > Thank you for making this widget!
> 
> I built this because I primarily use Claude Code (CLI), and VS Code already has usage info in the extension — I just wanted the same visibility on my bar YASB while coding. Glad it's useful for others too 😄
> 
> Thanks for building on top of it, the refresh button was exactly what was missing. Looking forward to getting this merged!

100%! And in my opinion, that's the best way to make software. To solve an issue you have of your own!

I think it fits perfectly in my setup:
image

ManaphatDev:
> > > > > Ah, much better
> > > > > image
> > > > 
> > > > 
> > > > Soo cool
> > > 
> > > 
> > > Thank you for making this widget!
> > 
> > 
> > I built this because I primarily use Claude Code (CLI), and VS Code already has usage info in the extension — I just wanted the same visibility on my bar YASB while coding. Glad it's useful for others too 😄
> > Thanks for building on top of it, the refresh button was exactly what was missing. Looking forward to getting this merged!
> 
> 100%! And in my opinion, that's the best way to make software. To solve an issue you have of your own!
> 
> I think it fits perfectly in my setup: image

Saw your setup — looks great! If you'd like a different icon for the widget, you can change it yourself: no code changes needed, it isn't hardcoded. The icon lives in the label (and label_alt) option, so you can swap it for any Nerd Font glyph, point it at your own image, or drop it entirely:


```yaml
# Nerd Font glyph
label: "<span>\udb80\ude7b</span> {five_hour}%"

# custom image
label: "<img src='.../claude.svg' width='14' height='14'> {five_hour}%"
```

jondmarien:
I just ran `claude -p ok` to refresh the invalidated token, and it took 4% of my token usage lol. I would love to brainstorm/find a way to re-validate this token other than forcing a prompt with claude code.

jondmarien:
Heads-up on a rename and two follow-up PRs that build on this one.

**Rename:** I retitled this PR to *"refresh button, expired-token indicator & configurable reset display"*. It started as just the refresh button, but with your review it grew to also cover the `{stale}` expired-token indicator and the per-window `reset_format` / `reset_show_date` display options — the old title undersold the scope, which would only get more confusing with more work layered on top.

**Why split the new work out:** rather than keep growing this PR, I kept the next features as separate, focused PRs so each stays reviewable (and so this one can merge on its own timeline). They're stacked on top of this branch:

- **#970 — local token-usage history.** An optional `Tokens` section in the popup with a Session/Today/Week/Month/Year toggle, the selected total, and an optional usage graph (reuses the shared `GraphWidget` from the CPU/Memory/GPU popups). Data comes from Claude Code's own local transcripts (`~/.claude/projects/**/*.jsonl`) — no API key, no network, numeric usage only (never message content), scanned incrementally off-thread. Also adds `{*_tokens}` bar placeholders. All off by default. (I evaluated the Admin API but it reports Console pay-per-token billing, which is ~empty on a Pro/Max subscription — local transcripts are the only source that reflects actual Claude Code usage.)
- **#971 — Claude API status indicator.** An optional green/yellow/orange/red dot (and an optional status line in the popup header) driven by the public status page (`status.claude.com/api/v2/status.json`, no auth). Adds `{status}`/`{status_text}` placeholders. Off by default.

**Stacking/review order:** #967 → #970 → #971. Because they're fork branches based on `main`, the #970/#971 diffs currently include this PR's commits; they'll drop out as each merges and I rebase. No rush — happy to have them reviewed in whatever order suits you, and to rebase once this lands.

jondmarien:
Added another optional bar-display option here (c808fc7): **`colorize_percent`** (default off). When enabled, the `{five_hour}`/`{seven_day}` percentage on the bar is coloured by usage level — wrap the percent in its own `<span class='percent'>` so only the number is tinted; it gets a `low`/`medium`/`high`/`critical` (or `unknown`) class.

While doing it I also **split the usage bands from 3 to 4** so the colour ramps green → yellow → orange → red:

| Band | Range | Colour |
|------|-------|--------|
| low | < 50% | green |
| medium | 50–64% | yellow |
| high | 65–79% | orange |
| critical | ≥ 80% | red |

This applies to **both** the popup progress bars and the new percentage text, so they stay consistent. Heads-up for theming: `.progress.high` (and `.percent.high`) is now **orange**, and the red band moved to the new `.critical` class — themes that want red at ≥ 80% should add `.critical`. Docs updated accordingly.

Happy to drop the 4th band back to 3 if you'd prefer to keep the existing low/medium/high behaviour unchanged — just say the word.

jondmarien:
Closing PR due to comment from @amnweb in PR #971. Remaking by hand.

PR #968, opened 2026-06-12
feat(burp_suite): add Burp Suite status widget

jondmarien:
## Summary

Adds a **Burp Suite** widget that shows, at a glance, whether [Burp Suite](https://portswigger.net/burp) is running --useful for pentesters/bug-bounty folks as a reminder that the proxy is up before browsing a target.

It reports one of three states, each with its own icon, status text, and CSS class:

- **offline** -- no Burp Suite window detected
- **running** -- Burp is running (REST API disabled, unreachable, or not yet ready)
- **ready** -- Burp is running *and* its REST API responds with `ready`

Detection is fully **read-only and needs no secrets**:
- Burp is found by enumerating top-level window titles (it always sets a `Burp Suite …` title), which also yields the edition (Pro / Community / Enterprise / DAST).
- The optional REST health check probes the service root `GET http://host:port/`, which returns `{"burp_status": "ready"}` and requires **no API key**. Set `rest_api.enabled: false` to do process-only detection.
- Window enumeration and the blocking HTTP probe run off the UI thread in a `QThread` (mirrors the OBS/Claude-usage service pattern); overlapping probes are deduplicated.

Supports `{icon}`/`{status}`/`{edition}` placeholders, `hide_when_offline`, a configurable poll interval, and `toggle_label` + `refresh` callbacks.

## Why window-title detection (design note)

I researched Burp's official REST API while building this. Its entire surface is `GET /` (health), `POST /scan`, `GET /scan/{task_id}`, and `GET /knowledge_base/issue_definitions` -- there's **no proxy/intercept endpoint and no "list scans" endpoint**, and scan progress is only queryable for tasks *you* started via the API. So richer "what is Burp doing right now" states aren't obtainable from a generic widget without false signals. Presence + REST-health is the honest, reliable scope; everything reported here is backed by the official API or the window manager.

## Files

- `src/core/widgets/yasb/burp_suite.py` -- widget
- `src/core/widgets/services/burp_suite/burp_client.py` -- detection + health-probe worker
- `src/core/validation/widgets/yasb/burp_suite.py` -- Pydantic config
- `docs/widgets/(Widget)-Burp-Suite.md` + README/sidebar entries

## Config example

```yaml
burp_suite:
  type: "yasb.burp_suite.BurpSuiteWidget"
  options:
    label: "<span>{icon}</span> {status}"
    label_alt: "<span>{icon}</span> Burp {edition}"
    update_interval: 5
    rest_api:
      enabled: true
      host: "127.0.0.1"
      port: 1337
    hide_when_offline: false
    callbacks:
      on_left: "toggle_label"
      on_middle: "refresh"
      on_right: "do_nothing"
```

## Testing

- `ruff check` / `ruff format` clean (Python 3.14).
- Verified in a 3.14 venv with the project deps: schema defaults + custom `model_validate` pass; `probe_burp` returns `offline` when Burp is absent, and `running`/`ready` with correct edition parsing when a Burp window title is present; live `win32gui.EnumWindows` enumeration confirmed working.

## Note on schema.json

I did **not** regenerate `schema.json` in this PR -- as that is your task, via the ci/cd pipeline. The new `BurpSuiteConfig` defs slot in cleanly; happy to regenerate it however you prefer, or leave it to the usual `chore(schema)` commit. Let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ choco install yasb
 | [Battery](https://github.com/amnweb/yasb/wiki/(Widget)-Battery) | Displays the current battery status. |
 | [Bluetooth](https://github.com/amnweb/yasb/wiki/(Widget)-Bluetooth) | Shows the current Bluetooth status and connected devices. |
 | [Brightness](https://github.com/amnweb/yasb/wiki/(Widget)-Brightness) | Displays and change the current brightness level. |
+| [Burp Suite](https://github.com/amnweb/yasb/wiki/(Widget)-Burp-Suite) | Shows whether Burp Suite is running, its edition, and REST API health. |
 | [Cava](https://github.com/amnweb/yasb/wiki/(Widget)-Cava) | Displays audio visualizer using Cava. |
 | [Copilot](https://github.com/amnweb/yasb/wiki/(Widget)-Copilot) | GitHub Copilot usage with a detailed menu showing statistics |
 | [CPU](https://github.com/amnweb/yasb/wiki/(Widget)-CPU) | Shows the current CPU usage and information. |
--- a/docs/_Sidebar.md
+++ b/docs/_Sidebar.md
@@ -15,6 +15,7 @@
     - [Battery](./(Widget)-Battery)
     - [Bluetooth](./(Widget)-Bluetooth)
     - [Brightness](./(Widget)-Brightness)
+    - [Burp Suite](./(Widget)-Burp-Suite)
     - [Cava](./(Widget)-Cava)
     - [Claude Usage](./(Widget)-Claude-Usage)
     - [Copilot](./(Widget)-Copilot)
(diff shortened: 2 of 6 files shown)


jondmarien:
For a bit of background on where this came from: the detection logic here grew out of [burpcord](https://github.com/jondmarien/burpcord), a Burp Suite extension I wrote that surfaces Burp activity as Discord Rich Presence. It went through PortSwigger's review and is now published on the official [BApp Store](https://portswigger.net/bappstore), so I've spent a fair amount of time inside Burp's extension APIs and its REST surface -- which is what informed the honest scoping of this widget (presence + REST health, since the official API genuinely can't back richer live states).

If there's appetite for it, I'd be more than happy to take a Burp/security integration to a larger scale within YASB -- e.g. a fuller popup menu (active scan tracking for API-initiated scans, issue counts, REST status detail) along the lines of the OBS or CPU/memory popups. Totally understand if presence-only is the right fit for core; just flagging that I'm keen to invest here if it's welcome.

PR #969, opened 2026-06-13
docs(claude-usage): mention <img> custom icon + list widget in README

ManaphatDev:
Two small docs-only fixes for the Claude Usage widget (shipped in v2.0.3):

1. **`` custom icon note** — in the widget's **Placeholders** section, document that the label can embed a custom image via an `` tag, alongside the existing Nerd Font `<span>` glyph example.
2. **README index** — the widget was never added to the README widget list; add it alphabetically between Cava and Copilot.

No code changes.


ManaphatDev:
Heads-up: this PR only updates documentation (README.md + wiki page). The failing ruff format check is coming from volume.py, which is unchanged from main.

I'm happy to clean that up in a follow-up PR if that's preferred.

I also have a popup screenshot matching the current behavior on main if you'd like me to add a preview image to the wiki page.

jondmarien:
Thanks @ManaphatDev. I've folded both of these into #973 and credited you via Co-Authored-By. They touch the same docs that PR rewrites, and keeping it to one PR is easier on the maintainer to review. The README list entry and the `` note are both in there, so this is safe to close once #973 lands.

PR #971, opened 2026-06-14
feat(claude_usage): Claude API status indicator

jondmarien:
## Summary

Adds an **optional** traffic-light **Claude API status** indicator to the widget — a green/yellow/orange/red dot on the bar and/or an status line in the popup header — so you can see at a glance if Claude is having issues.

It polls the public Claude status page (`https://status.claude.com/api/v2/status.json`, an Atlassian Statuspage — **no authentication**) off the UI thread via a refcounted shared `ClaudeStatusService` (keeps the last-known status if a fetch fails). The `status.indicator` (`none`/`minor`/`major`/`critical`) drives the dot colour.

## Highlights
- `{status}` bar placeholder — a dot coloured by level via dynamic `.status.<none|minor|major|critical|unknown>` classes (same technique as the Burp widget); `{status_text}` for the description.
- Optional status line (dot + description) in the popup header (`show_in_menu`).
- **Config** (`ClaudeStatusConfig`, off by default): `enabled`, `show_in_menu`, `icon`, `poll_interval`.
- Docs updated in `docs/`.

## Stacking
This is **stacked on #970** (which is stacked on #967), kept as a separate PR to keep each reviewable. The status-only changes are commits `6702719` and `f2327f1`; the earlier commits are #967 + #970 and drop out as those merge (I'll rebase). Easiest review order: #967 → #970 → this.

## Testing
- Offscreen Qt: `{status}` dot gets the right level class, menu status row updates, both features coexist.
- Live fetch confirmed (`none` → "All Systems Operational").
- `ruff check`/`format` clean. `schema.json` left for the maintainer chore commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #970, opened 2026-06-14
feat(claude_usage): local token-usage history (Session/Today/Week/Month/Year)

jondmarien:
## Summary

Adds an **optional** `token_history` feature to the Claude Usage widget: a **Tokens** section in the popup with a **Session / Today / Week / Month / Year** toggle, the selected period's total token count, and an optional usage graph — plus optional `{*_tokens}` bar placeholders. Inspired by [CodexBar](https://github.com/steipete/codexbar)'s usage view, but using the source that actually reflects Claude Code subscription usage.

The data comes from Claude Code's own local session transcripts (`~/.claude/projects/**/*.jsonl`) — **no API key, no network**. Only numeric token counts, timestamps, the model name and session id are read; **message content is never touched**. The scan is incremental (each file re-parsed only when its size/mtime changes) and runs off the UI thread in a refcounted shared service (`TokenHistoryService`), mirroring `ClaudeUsageService`.

### Why local JSONL (not the Admin API)
I evaluated Anthropic's Admin API (`/v1/organizations/cost_report` + `usage_report/messages`): it gives clean daily token/cost buckets but reports **Console pay-per-token billing**, which on a Pro/Max subscription is ~empty — it does not reflect Claude Code usage. The local transcripts are the only source that answers "how many tokens have I used in Claude Code," and they work for every user with no setup.

## Highlights
- **Tokens section**: period toggle, compact total (`15K`/`1.2M`/`218M`/`3.4B`/`1T`), reusing the existing menu styling.
- **Graph** (optional, `show_graph`): reuses the shared `GraphWidget` (the CPU/Memory/GPU chart). The window matches the selected period — Today/Session hourly, Week/Month daily, Year monthly.
- **Bar placeholders**: `{session_tokens}` `{today_tokens}` `{week_tokens}` `{month_tokens}` `{year_tokens}`.
- **Config** (`ClaudeTokenHistoryConfig`, all off by default): `enabled`, `default_period`, `show_graph`, `show_graph_grid`, `week_starts_on`, `count_cache_read`, `scan_interval`.
- Docs updated in `docs/`.

> **Note:** Session = the most-recent session's whole lifetime (can span days), so it may exceed Today. Totals are cache-read-heavy by default (cache reads dominate); set `count_cache_read: false` for "new work only."

## Stacking
[...]

PR #973, opened 2026-06-14
feat(claude_usage): refresh action, local token history and API status

jondmarien:
Following your comment on #971, I reworked this and went back through the contributing guidelines. This closes #967, #970 and #971 and puts everything in one PR.

Everything here is off by default and builds on the existing `claude_usage` widget. It was originally a stack of separate commits, one per feature, but per your request it's now squashed into a single commit.

### 1. Refresh, stale indicator, reset wording

A `refresh` callback plus a button in the popup header that force an immediate fetch and skip `cache_ttl`. Repeat fetches are still deduplicated while one is in flight, and a manual refresh rewrites the disk cache so the next scheduled tick reuses it instead of hitting the network again.

The header now uses the `stat_popup.py` layout you suggested (a `QFrame` header holding a `QLabel.text` title and a `QPushButton`), so the old ClickableLabel is gone. One thing to watch when theming: `.claude-usage-menu .header` is the container row now, and the title text is `.header .text`.

There's also a `{stale}` placeholder that shows a warning glyph once the OAuth token has expired. It reads the credentials' `expiresAt` only, never the token itself, and it's just a cue since the widget can't renew the token (running Claude Code does that). And two reset-format options, `five_hour_reset_format` / `seven_day_reset_format` with `reset_show_date`, which fix the broken "Resets in Sat 6:00 AM" line.

### 2. Local token-usage history

An optional Tokens section in the popup: a Session/Today/Week/Month/Year toggle, the total for the selected period, and an optional graph that reuses the shared `GraphWidget`. The same totals are on the bar through `{session_tokens}`, `{today_tokens}` and the rest.

It reads Claude Code's own session transcripts under `~/.claude/projects`, so there's no API key and no network call. Only the numeric token counts, timestamps, model name and session id get read; message content never does. A transcript is re-parsed only when its size or mtime changes, and the scan runs off the UI thread in a shared service like the usage poller.

### 3. Claude API status

An optional dot driven by the public status page at `status.claude.com` (no auth). `{status}` colours the dot by level through `.status.none` / `.minor` / `.major` / `.critical`, `{status_text}` gives the description, and you can show a status line in the popup header. If a poll fails it keeps the last known value rather than blanking out.

### 4. Per-model token breakdown

The Tokens section can now show where the tokens actually went. Turn on `token_history.show_models` and a short bar list appears (Opus 4.8, Sonnet 4.6, Fable 5, whatever else is in your transcripts), sorted by usage, top five. It follows the same period toggle as the rest of the section, so switching to Week or Month re-ranks the models for that window.

Same local data as the token history; nothing new gets read and there's still no network call. Model ids are shortened for display (`claude-opus-4-6` becomes `Opus 4.6`), with a fallback so an unrecognised id shows as-is instead of vanishing. I also moved the period toggle and running total down next to the graph, since the big number looked stranded once the model list sat between it and the chart.

### 5. Pinnable popup

A pin button in the popup header that keeps the popup open and lets you drag it. Originally built on the shared `PinnablePopup` in `stat_popup.py`; now runs on `PopupWidget`'s native `pinnable=True` instead (see section 9). The `pin_icon` / `unpin_icon` fields default to the same Segoe Fluent Icons glyphs the other widgets use; the button sits to the right of refresh and themes with `.claude-usage-menu .header .pin-btn`.

### 6. Fixed the popup flicker

Root cause was, on every refresh, `_refresh_usage_sections` was tearing down the 5h/7d `QFrame`s and inserting fresh ones, so Qt had to redo layout and briefly drew nothing before the new widgets landed - that's the flash. It's now `_sync_usage_sections`: the bar, reset line, percentage and date labels update in place on the same widgets, nothing gets destroyed on a normal poll.

### 7. Per-model weekly caps (e.g. Fable)

While digging into the flicker fix I noticed the usage endpoint now exposes model-scoped weekly caps (Opus, Sonnet, and a newer Fable one) through a `limits` array instead of the old flat `seven_day_opus`/`seven_day_sonnet` fields, which are null on current accounts. The widget reads that array generically off `scope.model.display_name`, so if your plan has one of these caps you get an extra bar in the popup for it, titled with the model's name, right after 7-Day. No config option needed - it shows up or not based on what the API actually returns, so a model added down the line needs no widget update either.

### 8. Popup built once and reused (persistent)

Following up on your `persistent` suggestion: the popup passes `persistent=True` to `PopupWidget` and is built once, the same pattern `control_center` uses. Opening it again just shows/repositions the existing window instead of reconstructing the whole widget tree; a `_show_menu` helper pushes the latest data in right after `show()` so nothing stale is visible from before it was hidden.

While testing that, I also caught a real resize bug: switching from a period with several models (Month/Year) to one with fewer (Session/Today) left a gap under the graph instead of shrinking to fit. `sizeHint()` was reading a cached value from before the model rows were removed - an earlier version of this code called `_menu_layout.activate()` to force that recompute before resizing, and it had gotten dropped along the way. Restored it, so the popup now shrinks correctly in both directions.

### 9. Moved to `PopupWidget.pinnable`, reverted the resize hack, restyled the docs

Three follow-ups from the last review round:

- **`stat_popup.py`**: reverted the `PinnablePopup.resizeEvent` override I'd added for a pinned-popup resize/snap issue. You couldn't reproduce it after the popup went persistent, and neither could I, so it's gone.
- **`claude_usage.py`**: dropped `PinnablePopup`/`create_pin_button` entirely. `_build_menu` now builds a plain `PopupWidget(persistent=True, pinnable=True)`, and the pin button calls `menu.set_pinned(checked)` on toggle - followed the `Writing-Widget.md #9` example closely. `cpu`/`memory`/`gpu` are untouched; you mentioned moving those yourself later.
- **Docs**: rewrote the example style to match the Windows 11 look `control_center` already ships instead of inventing new colours - same `rgba(32, 32, 32, 0.6)` panel, same `#47AFF5` accent, same thin 1px separators between sections instead of boxed cards. Not a pixel match to the Bluetooth screenshot you attached, but consistent with what's already in the app.

### Notes

Docs are updated under `docs/`, including a full copy-pasteable style example and a screenshot of the popup under "Example Widget". Ruff check and format are clean. I tested against my own install plus offscreen Qt: the refresh button and callback, the stale glyph, both reset formats, the token section across all five periods, the per-model breakdown and its re-ranking when switching periods, the pin button (now on native `PopupWidget.pinnable`), the status dot at each level, the flicker fix, the per-model weekly bar, and the persistent popup (opened/closed it repeatedly and switched between periods with wildly different model counts, pinned and unpinned, watching for any resize/jump/gap).

This also folds in the docs-only changes from #969 by @ManaphatDev (the custom `` icon note and the README index entry), credited via Co-Authored-By. They touch the same docs this PR rewrites, and #969 was stuck on an unrelated CI failure, so it was cleaner to bring them in here.

### DIFF
--- a/docs/widgets/(Widget)-Claude-Usage.md
+++ b/docs/widgets/(Widget)-Claude-Usage.md
@@ -2,7 +2,9 @@
 
 Shows your Claude (Claude Code) subscription usage on the bar - the 5-hour and 7-day
 rolling limit utilization - with a popup menu that displays both windows as progress bars
-together with their reset times (a countdown plus the exact reset timestamp).
+together with their reset times (a countdown plus the exact reset timestamp). Any per-model
+weekly cap your plan has (e.g. a separate Fable limit) shows up as an extra bar automatically;
+see [Per-model weekly caps](#per-model-weekly-caps).
 
 The data is read from the same OAuth credentials the Claude Code CLI uses
 (`~/.claude/.credentials.json`) and fetched from Anthropic's usage endpoint. No API key or
@@ -14,21 +16,35 @@ extra configuration is required as long as you are signed in to Claude Code.
 | `label_alt`       | string  | `'Claude {seven_day}%'` | The alternative format string, toggled by the `toggle_label` callback. |
 | `update_interval` | integer | `60` | How often the label and reset countdown are refreshed, in seconds. Must be between 30 and 3600. |
 | `cache_ttl`       | integer | `120` | How long (seconds) a fetched result is cached on disk before the endpoint is queried again. The endpoint is rate-limited, so keep this at a sane value. |
+| `five_hour_reset_format` | string | `'relative'` | How the 5-hour window's reset line is phrased in the popup: `relative` (`Resets in 4h 11m`) or `absolute` (`Resets on Sat @ 6:00 AM`). |
+| `seven_day_reset_format` | string | `'absolute'` | How the 7-day window's reset line is phrased in the popup: `relative` or `absolute`. |
+| `reset_show_date` | boolean | `true` | In `absolute` mode, include the month/day (`Resets on Sat, Jun 13 @ 6:00 AM`) so two windows resetting on the same weekday stay distinguishable. |
+| `token_history`   | dict    | `{'enabled': false, ...}` | Optional local token-usage history. See [Token history](#token-history). |
+| `status`          | dict    | `{'enabled': false, ...}` | Optional Claude API status indicator. See [API status](#api-status). |
 | `tooltip`         | boolean | `true` | Whether to show a summary tooltip on hover. |
 | `callbacks`       | dict    | `{'on_left': 'toggle_menu', 'on_middle': 'do_nothing', 'on_right': 'toggle_label'}` | Mouse-click callbacks. |
-| `menu`            | dict    | `{'blur': true, 'round_corners': true, 'round_corners_type': 'normal', 'border_color': 'System', 'alignment': 'right', 'direction': 'down', 'offset_top': 6, 'offset_left': 0}` | Popup menu settings. |
+| `menu`            | dict    | `{'blur': true, 'round_corners': true, 'round_corners_type': 'normal', 'border_color': 'System', 'alignment': 'right', 'direction': 'down', 'offset_top': 6, 'offset_left': 0, 'pin_icon': '', 'unpin_icon': ''}` | Popup menu settings. |
 
 ## Placeholders
 
 The label is plain text by default. You can prepend a Nerd Font glyph in a `<span>` if you
-want an icon (e.g. `<span>\U000f06a9</span> {five_hour}%`). The following placeholders can be
-used in `label` / `label_alt`:
+want an icon (e.g. `<span>\U000f06a9</span> {five_hour}%`), or embed your own image with an
+`<img>` tag (e.g. `<img src='C:/path/to/claude.svg' width='14' height='14'> {five_hour}%`). The
+following placeholders can be used in `label` / `label_alt`:
 
 - `{five_hour}` - 5-hour window utilization (percent, `--` when unavailable).
 - `{seven_day}` - 7-day window utilization (percent, `--` when unavailable).
 - `{five_hour_reset}` - time until the 5-hour window resets. Shown as a countdown when under a
   day away (e.g. `4h 27m`), otherwise as a local weekday + time (e.g. `Sat 6:00 AM`).
 - `{seven_day_reset}` - time until the 7-day window resets (e.g. `Sat 6:00 AM`).
+- `{stale}` - a warning glyph shown only while Claude Code's OAuth token has expired, empty
+  otherwise. Place it in its own `<span>` (e.g. `{five_hour}% <span class='stale'>{stale}</span>`).
+- `{session_tokens}` `{today_tokens}` `{week_tokens}` `{month_tokens}` `{year_tokens}` - compact
+  token totals (e.g. `1.2M`) for each period. Require `token_history.enabled`; `--` otherwise.
+- `{status}` - a status dot, coloured by the current Claude API status level via
+  `.status.<none|minor|major|critical>` classes. Place it in its own `<span>`. Requires
+  `status.enabled`; empty otherwise.
+- `{status_text}` - the status description (e.g. `All Systems Operational`).
 
 ```yaml
 claude_usage:
@@ -40,7 +56,7 @@ claude_usage:
     cache_ttl: 120
     callbacks:
       on_left: "toggle_menu"    # open the usage menu
-      on_middle: "do_nothing"
+      on_middle: "refresh"      # force an immediate re-fetch, bypassing cache_ttl
       on_right: "toggle_label"  # switch the bar text between 5h and 7d
     menu:
       blur: true
@@ -59,8 +75,10 @@ claude_usage:
 - **label_alt:** The alternative format string, toggled with the `toggle_label` callback.
 - **update_interval:** How often the bar label and reset countdown are refreshed, in seconds (30-3600).
 - **cache_ttl:** How long a fetched result is cached on disk before the usage endpoint is queried again. Because the endpoint is rate-limited (HTTP 429), the widget serves the last cached value on any error instead of going blank.
+- **five_hour_reset_format / seven_day_reset_format:** How each window's reset line is phrased in the popup. `relative` shows a countdown (`Resets in 4h 11m`); `absolute` shows a local weekday and time (`Resets on Sat @ 6:00 AM`). The exact reset timestamp is always shown on the line below.
+- **reset_show_date:** In `absolute` mode, include the month/day in the reset line so the 5-hour and 7-day windows can be told apart when they fall on the same weekday. No effect in `relative` mode.
 - **tooltip:** Whether to show a summary tooltip on hover.
-- **callbacks:** Mouse-click callbacks. Built-in actions: `toggle_menu` (open/close the popup menu), `toggle_label` (swap between `label` and `label_alt`), `do_nothing`, and `exec`.
+- **callbacks:** Mouse-click callbacks. Built-in actions: `toggle_menu` (open/close the popup menu), `toggle_label` (swap between `label` and `label_alt`), `refresh` (force an immediate re-fetch, bypassing `cache_ttl`), `do_nothing`, and `exec`.
 - **menu:** A dictionary specifying the popup menu settings:
   - **blur:** Enable blur effect for the menu.
   - **round_corners:** Enable round corners (not supported on Windows 10).
@@ -69,6 +87,7 @@ claude_usage:
   - **alignment:** Horizontal alignment of the menu (`left`, `right`, `center`).
   - **direction:** Whether the menu opens `down` or `up`.
   - **offset_top / offset_left:** Pixel offsets for fine positioning.
+  - **pin_icon / unpin_icon:** Nerd Font glyphs for the pin button in the popup header. The button keeps the popup open and lets it be dragged when pinned.
 
 ## Authentication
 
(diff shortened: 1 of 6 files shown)


jondmarien:
Pushed two more commits on top (the last two): a per-model token breakdown in the Tokens section, and a pinnable popup that matches the cpu/memory/gpu pin button. Both are off by default with no new deps or network calls; the breakdown reuses the transcript data the token-history scan already reads, and the pin reuses the shared `PinnablePopup`. The single-sample graph case is handled in the widget itself, so the shared `GraphWidget` is untouched. Description updated with sections 4 and 5.

amnweb:
Can you show me a screenshot of how this widget looks now? There have been so many changes. Also, please squash these commits into a single commit.

jondmarien:
> Can you show me a screenshot of how this widget looks now? There have been so many changes. Also, please squash these commits into a single commit.

Yes, 100%. I will also squash these into a single commit. I had them separated so you could read them, and have no problem squashing them.

Here is what the widget looks like now:
Full View:
image
Hourly:
image
Weekly:
image

amnweb:
Is this a bug in the progress bar, or have you set padding or margin in the style? It does not fit the container, as you can see in the screenshot.
image

jondmarien:
> Is this a bug in the progress bar, or have you set padding or margin in the style? It does not fit the container, as you can see in the screenshot. image

Let me confirm for you, i hadn't noticed that. Good eyes!!

amnweb:
I just tested the Copilot widget, and everything looks fine there.

image

jondmarien:
> I just tested the Copilot widget, and everything looks fine there.
> image

okay yeah it must be the claude widget then. I took inspo from the copilot widget for this. Let me fix this!

jondmarien:
this also affects the model bars, not just the hourly/weekly bars. will fix for all in this widget!

edit: I think because I implemented my own `UsageBar`, and did not reuse the `ProgressBar` from Copilot, it created some differences. fixing now.

edit2: i think its my css 🙈 let me investigate further

jondmarien:
@amnweb could you provide your css for the copilot widget regarding the `.progress/.fill/.section` rules so I can confirm it is a css issue? I am almost 100% sure, but I am not the best with css.

jondmarien:
Squashed the branch into a single commit, as you asked.

What's folded into it:
- refresh action plus a `{stale}` indicator that fires once the OAuth token expires, and per-window reset wording that fixes the old "Resets in Sat 6:00 AM" line
- local token-usage history (Session/Today/Week/Month/Year) read from Claude Code's own transcripts under `~/.claude/projects`, scanned off-thread and incrementally; the cache is pruned on each scan (hourly past 15 days, daily past ~13 months) so it stays bounded
- the Claude API status dot from the public status page, no auth
- a per-model token breakdown in the Tokens section that follows the period toggle; model names are derived from the id so new models need no upkeep, and fast-mode usage gets its own row
- the pinnable, draggable popup
- review fixes from @ManaphatDev (credited as co-author) and two small follow-ups: the popup now resizes to fit the selected period, and the token formatter rounds correctly at the unit boundaries (999,999 reads as 1M instead of 1000K)

On the bar that didn't fit its container: see comment below

jondmarien:
> @amnweb could you provide your css for the copilot widget regarding the `.progress/.fill/.section` rules so I can confirm it is a css issue? I am almost 100% sure, but I am not the best with css.

nevermind, i had max-height set for some reason. 

looks like this now:
image

jondmarien:
the only thing I noticed is this placement of the tokens per model usage is not inline. would you like me to fix this as well?

image

amnweb:
But this is a problem with the progress bar. You should make it the same size, but I have no idea how because this popup window can be changed via CSS, so you can't limit size there :(

image

jondmarien:
> But this is a problem with the progress bar. You should make it the same size, but I have no idea how because this popup window can be changed via CSS, so you can't limit size there :(
> image

hmmm.... really good point amnweb....i will take a bit to think about this. will comment when I have a working idea.

jondmarien:
Fixed...implemented a `QGridLayout`! Will make a final comment on this. 

image

jondmarien:
Pushed the bar-alignment fix (still one commit). You were right that the bars came out different sizes (I also talked with @ManaphatDev about this); the cause was the name and total labels being content-width, so the stretchy bar column started and ended at a different x on each row. I moved the per-model rows into a `QGridLayout`: Qt sizes the name and total columns to their widest cell at layout time, so the bar column is identical on every row and the fills scale against the same track. No manual width math, so the font or popup width can't throw it off.

Caught one more while testing: a pinned popup snapped back to its spawn point on a refresh tick. `PopupWidget.resizeEvent` re-anchors to the bar on every resize, and this widget resizes when the model count changes between periods. `PinnablePopup` now keeps the dragged position while pinned (cpu/memory/gpu avoid this by never resizing on tick; `claude_usage` does, so it needed handling).

Verified locally across all five periods, on first open and after refresh, pinned and unpinned.

Final look:
image

amnweb:
This should probably be done in CSS. Set a minimum width for the left and right labels, and let the progress bar stretch to fill the remaining space, it was more simpler... 

May I ask what the reason was for creating a new `create_pin_button` function?

jondmarien:
> This should probably be done in CSS. Set a minimum width for the left and right labels, and let the progress bar stretch to fill the remaining space, it was more simpler...
> 
> May I ask what the reason was for creating a new `create_pin_button` function?

sure, i can revert it. I had some issues when doing it with CSS:
image
But I'll try it again.

I created the `create_pin_button` function due to that being re-created multiple times over the codebase, as a helper. If you would like it to be removed, I have no issue doing so, just let me know.

amnweb:
> I created the `create_pin_button` function due to that being re-created multiple times over the codebase, as a helper. If you would like it to be removed, I have no issue doing so, just let me know.

Where is `create_pin_button` used multiple times?
I have to test this PR somehow, I can't trust AI sorry, there are so many changes.

jondmarien:
> > I created the `create_pin_button` function due to that being re-created multiple times over the codebase, as a helper. If you would like it to be removed, I have no issue doing so, just let me know.
> 
> Where is `create_pin_button` used multiple times? I have to test this PR somehow, I can't trust AI sorry, there are so many changes.

You should be able to test it normally, however you do so. I followed all the contributing guidelines, and I followed the structure of the rest of the codebase, and how other widgets are written. So since there are many changes, it must be AI? I walked through every single changed line, I can help you test it if you need. 

I also noticed there are no pytests in the codebase, so I did not write any, but I did run off-screen tests with expected values, with many iterations, and as far as I can tell, everything is good. @ManaphatDev also has been testing this PR as well. 

For `create_pin_button`, it is used in the memory/cpu/gpu widgets, as well as the `claude_usage` widget:
- `build_stat_popup` at [stat_popup.py:267]; the shared popup builder for the cpu/memory/gpu widgets.
- `claude_usage._build_menu` at [claude_usage.py:649]; claude widget.

I take pride and care in what I submit to PRs, granted, the first few were messy, but I took time with this one, trying to adhere to all of your requests. I would love to have this PR merged within YASB, I use your software often, and contributing to OSS is always fun.

amnweb:
Sorry for the delay. No, I can't test anything there. I need to mock up data. I mean, I can test, but without data it's not worth it. 
But I noticed you missed some elements in the docs style example? If you can, please make a full example so users can just copy/paste the style.

About this `resizeEvent`, I'm not sure what the problem is. Does `updateGeometry()` , `ensurePolished()` or `adjustSize()` ... or something simple like that not work?

Edit: About `build_stat_popup`, yeah this is a shared module, and once it's created, Python caches all of it. So nah, you don't need a helper for this, but ok, it's fine.

jondmarien:
> Sorry for the delay. No, I can't test anything there. I need to mock up data. I mean, I can test, but without data it's not worth it. 

I mocked the data locally, as well as testing with my `.claude` files. I understand you do not use Claude, which is fine, but how can we get this PR moving forward if you cannot test it without real data? I can look into sharing some of the files with you, as long as they do not contain any PII or other data I would prefer not to be shared.

> About this `resizeEvent`, I'm not sure what the problem is. Does `updateGeometry()` , `ensurePolished()` or `adjustSize()` ... or something simple like that not work?

For the resizing issue, I am also not sure. I will try the suggested methods. 

> Edit: About `build_stat_popup`, yeah this is a shared module, and once it's created, Python caches all of it. So nah, you don't need a helper for this, but ok, it's fine.

For the helper due to the "repeated usage" (barely) of the `create_pin_button`, just make your decision whether you would like it in the codebase and I can follow suit. 

> But I noticed you missed some elements in the docs style example? If you can, please make a full example so users can just copy/paste the style.

Finally, regarding this, I can 100% fill out the missing styles. Apologies for missing them in the first place!

amnweb:
It's fine if this works for you. I won't waste time with fake data. Just make sure the docs are fine so we're good.

jondmarien:
> It's fine if this works for you. I won't waste time with fake data. Just make sure the docs are fine so we're good.

Alright, thanks amnweb. I'll fix the docs up. Look out for this in the next few days, just dealing with life stuff right now.

amnweb:
OK I had some time and tested this. What's the main problem here? Instead of keeping the existing progress bars and labels and just updating their text or values in place, the code completely destroys the old UI frames with `old.deleteLater()` and inserts brand new ones every single time the data refreshes. 
When you delete and re-insert widgets, Qt has to recalculate the layout and briefly shows an empty space before drawing the new widgets. This causes a visual flicker or flash.

_By the way, that's an example style from the docs._


https://github.com/user-attachments/assets/8610cce4-34aa-4bc7-8c23-e51abeb6e1f0

jondmarien:
Found it. On every refresh, `_refresh_usage_sections` was destroying the 5h/7d `QFrame`s and inserting new ones in their place, so Qt had to redo the layout and briefly drew nothing before the new widgets landed -- that's the flash you were seeing. Fixed by mutating the existing bar, reset line, percentage and date labels in place instead (`_sync_usage_sections`), so a normal poll doesn't touch the widget tree at all anymore. The frames only get built once, when the popup opens.

Also, while I was in there: turns out Anthropic added model-scoped weekly caps to the usage endpoint -- Opus and Sonnet have had this for a while apparently, and there's a newer one for Fable. The old `seven_day_opus`/`seven_day_sonnet` fields are null now; the real data lives in a `limits` array with a `scope.model.display_name` field. I wired that up generically, so if your plan has one of these caps it just shows up as an extra bar after 7-Day, titled with whatever model name the API gives back. No new config option, nothing hardcoded to Fable specifically -- whatever the endpoint reports is what renders.

Squashed everything back into one commit like you asked, docs are filled out (including a full copy/paste style example and a screenshot of the popup), and I re-tested the whole thing end to end -- both reset formats, all five token periods, the model breakdown, pin button, status dot, and the flicker fix specifically (pinned + open, mashing refresh).

Let me know if there's anything else you want changed before this goes in.

Final Widget Preview (same one attached in wiki for Claude Widget):
image

amnweb:
I've added a `persistent` option to the `PopupWidget`, which prevents it from being deleted. This should fix all your issues with resizing, jumping, speed, and so on. For example, check `https://github.com/amnweb/yasb/blob/main/src/core/widgets/yasb/control_center.py` and look at the `PopupWidget` class itself. I've started moving some popups to use persistent because it didn't delete the layout, and it works faster and without the problems we had before.

jondmarien:
> I've added a `persistent` option to the `PopupWidget`, which prevents it from being deleted. This should fix all your issues with resizing, jumping, speed, and so on. For example, check `https://github.com/amnweb/yasb/blob/main/src/core/widgets/yasb/control_center.py` and look at the `PopupWidget` class itself. I've started moving some popups to use persistent because it didn't delete the layout, and it works faster and without the problems we had before.

Awesome! Let me move over to this `persistent` option on the `PopupWidget`. Sounds like a great change. Thanks amnweb.

jondmarien:
I also made sure to update the docs to match the repo-wide grammar style commit that was recently pushed.

jondmarien:
Went with the `persistent` option, thanks for pointing that out. The popup now gets built once and reused - `_show_menu` just shows/repositions the existing window on every open instead of reconstructing the widget tree, same as `control_center`. Should take care of the resizing, jumping and speed you mentioned.

While I was testing that I found a real bug on top of it: switching from a period with a bunch of models (Month/Year) down to one with fewer (Session/Today) left a gap under the graph instead of shrinking to fit. Turned out `sizeHint()` was reading a stale cached value from before the model rows got removed - `_menu_layout.activate()` used to force that recompute before resizing and it had gotten dropped somewhere along the way. Restored it, so the popup now sizes correctly switching either direction.

Squashed back into one commit and pushed. Let me know how this looks on your end.

amnweb:
I just tested this widget again, and it looks fine to me. I don't think you need to change `src\core\utils\stat_popup.py` and mess with `resizeEvent`. If it's possible to revert the stats_popup change, that would be nice. I can't find any problem with updating data without this resizing, I watched it in the pinned window with changes in real time and it looks fine to me. If I need to know something about that or to reproduce the problem, let me know. Also, it would be nice to update this style because it doesn't look very good. I mean the docs and example style. Also, if possible, matching Windows 11 UI in the default CSS would be nice, because in the last few months we have moved almost every widget to default to match Windows 11 UI.

Example
image

_Note: I'm just suggesting about the style, if it's not possible to match this UI, just update the style so it looks right._

amnweb:
Ok @jondmarien I've added support for pin options in the current popup widget, so now you can use it. Here's an example and docs: https://github.com/amnweb/yasb/blob/main/docs/Writing-Widget.md#9-using-popupwidget-for-dropdown-menus

I'll also refactor other widgets and move from `stat_popup.py` to `PopupWidget` when I have time, so we can have a clear codebase here.

Usage is very simple: just add support in `PopupWidget` to support a pin (`pinnable: bool`), and then toggle it inside the widget menu by calling `menu.set_pinned(bool)`. Anyway, it's all [documented](https://github.com/amnweb/yasb/blob/main/docs/Writing-Widget.md#9-using-popupwidget-for-dropdown-menus).

jondmarien:
> Ok @jondmarien I've added support for pin options in the current popup widget, so now you can use it. Here's an example and docs: https://github.com/amnweb/yasb/blob/main/docs/Writing-Widget.md#9-using-popupwidget-for-dropdown-menus
> 
> I'll also refactor other widgets and move from `stat_popup.py` to `PopupWidget` when I have time, so we can have a clear codebase here.
> 
> Usage is very simple: just add support in `PopupWidget` to support a pin (`pinnable: bool`), and then toggle it inside the widget menu by calling `menu.set_pinned(bool)`. Anyway, it's all [documented](https://github.com/amnweb/yasb/blob/main/docs/Writing-Widget.md#9-using-popupwidget-for-dropdown-menus).

Hi @amnweb, sounds good, I will get this done in the coming days, possibly tonight. I'll look through the added documentation, thanks for that :)

jondmarien:
> I just tested this widget again, and it looks fine to me. I don't think you need to change `src\core\utils\stat_popup.py` and mess with `resizeEvent`. If it's possible to revert the stats_popup change, that would be nice. I can't find any problem with updating data without this resizing, I watched it in the pinned window with changes in real time and it looks fine to me. If I need to know something about that or to reproduce the problem, let me know. Also, it would be nice to update this style because it doesn't look very good. I mean the docs and example style. Also, if possible, matching Windows 11 UI in the default CSS would be nice, because in the last few months we have moved almost every widget to default to match Windows 11 UI.
> 
> Example
> image
> 
> _Note: I'm just suggesting about the style, if it's not possible to match this UI, just update the style so it looks right._

I will also match the UI styling default to be the Windows 11 UI, and revert the stats_popup changes.

amnweb:
> Hi @amnweb, sounds good, I will get this done in the coming days, possibly tonight. I'll look through the added documentation, thanks for that :)

When you have time, no problem. The new preview is coming tonight anyway. We made all of this inside this PR over complicated, so moving all this in the PopupWidget is the right way.

jondmarien:
Reverted the `stat_popup.py` resizeEvent override, as you suggested.

Also moved this widget off `PinnablePopup` entirely. `_build_menu` now constructs a plain `PopupWidget(persistent=True, pinnable=True)`, and the pin button just calls `menu.set_pinned(checked)` on toggle -- followed the Writing-Widget.md #9 example pretty much line for line. Left `cpu`/`memory`/`gpu` on `stat_popup.py` since you mentioned migrating those yourself later.

Docs got a pass as well; Pulled `control_center`'s actual values -- the `rgba(32, 32, 32, 0.6)` panel, the `#47AFF5` accent, the thin 1px separators between sections instead of the old boxed-card look -- so the example style matches what's already shipping elsewhere in the app. Not a pixel-for-pixel match to the Bluetooth screenshot, but I tried to match the same design language.  

Squashed back into one commit and pushed. Let me know if anything's still off.

amnweb:
Thanks for the update, everything works fine for me!
If that's how it should look, we're good to merge once you're done.
image

Just a couple of quick things before we wrap up:

`README.md` Please leave this out for now, I'll update it myself once the stable release lands.
`stats_popup.py` Could you please revert any changes here? No modifications are needed in this file since I'm planning to handle widget extractions separately later on.

jondmarien:
Done on both. Dropped the README.md line entirely, that'll wait for you to add once stable ships. And `stat_popup.py` is back to stock, byte for byte against main - no diff at all now. Doesn't cost anything since `claude_usage` already moved off `PinnablePopup` onto the native `PopupWidget.pinnable` path last round, so nothing there depended on the old helper anymore anyway.

Squashed and pushed again.

amnweb:
Thanks.

jondmarien:
Thanks for the merge, @amnweb. I appreciate you working with me on this, and I apologize for the rocky start. Hope to contribute more to YASB in the future.

Issue #976, opened 2026-06-16
CPU Widget Does Not Identify CPU Usage Correctly


TRGameLore:
### What is the request?

When I game, I run into the issue where the CPU widget doesn't display the actual usage. It only displays half of the actual usage. The base speed of my CPU is 2.50 GHz, but it gets turbo'd to ~4.50 GHz. It would be nice if it could be added so it reads correctly all the time. Image included as an example (CPU widget: 25.9% | Actual usage: 52%). 

<img width="1902" height="807" alt="Image" src="https://github.com/user-attachments/assets/141d0879-9e90-4920-a197-10fcb4bc537a" />

### Why is it useful?

I love the fact that I can monitor my PC's usage with the top bar. But It's extremely inconvenient looking at it while I'm playing a game, seeing that I'm only at 45% usage. When in reality I am at 90% usage.

TRGameLore:
1. Windows 10
2. YASB v.2.0.4
3. YASB Log

[yasb.log](https://github.com/user-attachments/files/28985656/yasb.log)

config.yaml code:
  cpu:
    type: "yasb.cpu.CpuWidget"
    options:
      label: "<span>\uF4BC</span> {info[percent][total]}%"
      label_alt: "<span>\uF4BC</span> {info[percent][total]}%"
      update_interval: 2000
      histogram_icons:
        - '\u2581' # 0%
        - '\u2581' # 10%
        - '\u2582' # 20%
        - '\u2583' # 30%
        - '\u2584' # 40%
        - '\u2585' # 50%
        - '\u2586' # 60%
        - '\u2587' # 70%
        - '\u2588' # 80%+
      histogram_num_columns: 8
      callbacks:
        on_right: "exec cmd /c Taskmgr"
      container_padding:
        top: 0
        left: 0
        bottom: 0
        right: 0

It varies whether it shows ~half CPU usage; typically, when I am doing something resource-intensive, such as gaming.

YASB shows the same CPU usage when I am idle or not gaming.

amnweb:
@TRGameLore which CPU do you have there?

About Turbo/Boost, YASB works a little differently than Windows, especially on Windows 11 where CPU usage can go above 100% when boost is active. In YASB, the calculation is always capped at 100%.

**Note: There is a Libre Hardware Monitor widget that might work better for your CPU.**

TRGameLore:
[...]

PR #977, opened 2026-06-17
fix(cpu): use correct pdh counters depending on windows version

amnweb:
- use % Processor Time on Windows 11 24H2+ (build 26100+)
- use % Processor Utility with fallback to % Processor Time on older Windows versions
- apply build-dependent selection to both total and per-core counters


amnweb:
I need more feedback. 
This PR is related to issue https://github.com/amnweb/yasb/issues/976 where Windows 10 reports wrong CPU usage.

PR #998, opened 2026-06-30
feat(scroll): add invert_wheel option

notTamion:
This is useful for laptop trackpads (when scrolling with 2 fingers on the pad) as moving your fingers up actually scrolls down however i think its more intuitive to move fingers up to increase volume. 

I can implement this for all the other widgets that have a system like this (brightness, mic i think?) but i would like to know whether this is something you want before i put further work in this. if so please give me a list of all widgets that have something like this (since i am not that familiar with the widgets other than the ones i use)

the slider in the drop down menu doesn't have their behaivour changed in the pr since qt widgets only supports inverting all inputs, so also arrow keys which we don't want to invert. however that can be added in the future if that change is PRd to QT (i might do that if i have some time)

### DIFF
--- a/docs/widgets/(Widget)-Brightness.md
+++ b/docs/widgets/(Widget)-Brightness.md
@@ -8,6 +8,7 @@ Displays your screen's brightness level and lets you adjust it on the fly. You c
 | `label_alt`     | string  | `"Brightness {percent}%"`                                          | The alternative format string for the brightness widget. |
 | `tooltip`  | boolean  | `True`        | Whether to show the tooltip on hover. |
 | `scroll_step`   | integer | `1`                                                                       | The step size for scrolling the brightness level. This value must be between 1 and 100. |
+| `invert_wheel`     | boolean     | `false`                  | Whether to invert wheel scroll direction (useful on laptop trackpads) |
 | `brightness_icons` | list  | `['\udb80\udcde', '\udb80\udcdd', '\udb80\udcdf', '\udb80\udce0']`                    | A list of icons representing different brightness levels. The icons are used based on the current brightness percentage. |
 | `hide_unsupported` | boolean | `True` | Whether to hide the widget if the current system does not support brightness control. |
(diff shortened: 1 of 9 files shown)


amnweb:
Well, this is scrolling directly on the bar, right? So I think only the microphone widget has it. For the brightness widget, I think it's a little bit different logic there because we have multiple monitor support and etc., so you don't need to waste time there. But the microphone shares the same logic as volume, so you can add it there.

notTamion:
not the bar (thats what i talked about with the qt widget stuff). rn whats being changed is only the scrolling on the widget directly 
image

amnweb:
> not the bar (thats what i talked about with the qt widget stuff). rn whats being changed is only the scrolling on the widget directly image

When you hover your mouse over the widget in the bar and scroll, that's what I mean when I say "directly on the bar." Is that what you mean?

notTamion:
Ah yeah i thought you were talking about this bar:
image

but alright then. once i have time i will implement it for microphone and ig im gonna check out brightness just to see.

notTamion:
Ready for review

notTamion:
went ahead and fixed another issue i found where when using the qt bar and sliding it to mute or unmute it wont set the mute (let me know if you would rather have this be a seperate pr then i will create a new one for that)

amnweb:
It works fine for me, or do you mean it doesn't set mute when you use `QSlider` and popup menu?

https://github.com/user-attachments/assets/f51c5276-5402-4ea3-9fd9-be74617ca937

notTamion:
> It works fine for me, or do you mean it doesn't set mute when you use `QSlider` and popup menu?
> 
> https://github.com/user-attachments/assets/f51c5276-5402-4ea3-9fd9-be74617ca937
> 
> 

yeah thats what i meant

amnweb:
Well, you can do it in the same PR. I believe it's a small change.

notTamion:
> Well, you can do it in the same PR. I believe it's a small change.

already in

PR #1054, opened 2026-08-06
feat(taskbar): add CSS classes for minimized, and visible windows

PSB1234:
Added CSS classes to distinguish visible background windows from minimized Taskbar windows.

- Visible, non-focused windows receive `.not-focused`
- Minimized windows receive `.minimized`
- Existing `.running`, `.foreground`, and `.flashing` classes remains as it is

Previously, visible background windows and minimized windows shared the same **.running** class, making it impossible to style them differently. This lets user to further control the look of taskbar icons. **.running** has become a superset of **.minimized** and **.not-focused**.

## Example

```css
.taskbar-widget .app-container.minimized {
    opacity: 0.4;
    background-color: transparent;
}
.taskbar-widget .app-container.not-focused {
    background-color: rgba(255, 255, 255, 0.25);
}
```
image

Here, In the above image, File explorer has cyan grayish background and vscode has gray background while other running apps which are minimized, grayed out.

### DIFF
--- a/docs/widgets/(Widget)-Taskbar.md
+++ b/docs/widgets/(Widget)-Taskbar.md
@@ -75,6 +75,7 @@ taskbar:
 > Note:
 > When **preview** is enabled **tooltip** are automatically disabled to avoid overlap.
 
+
 ## Style
 ```css
 .taskbar-widget {} /* Main container for the taskbar widget */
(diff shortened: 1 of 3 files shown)


amnweb:
`.not-focused`? Isn't that the same as the app container `.taskbar-widget .app-container`?

PSB1234:
no, `.app-container` is the base class/style applied to every taskbar application button, regardless of state.  
`.foreground `identifies the focused application.` .running` is the base state for non-focused applications.  
`.not-focused` and `.minimized `are more specific state classes for non-focused apps: `.not-focused` means the app is visible on screen but not focused, while `.minimized` means the app is minimized. This lets users style visible background apps differently from minimized apps.
image
In this above example, If I want to style css for all applications in taskbar widget, I will use `.app-container`. If I want to change css for how main focused window look, I will use `.app-container.foreground` specifying I wanna apply this state to only main application, In example, I used green color. If I want to apply css to all background/not-focused application. I will use `.running` but, here we can't specify which state of window the not-focused applications are in, Are they on-screen (like the orange color in example image also this doesn't include minimized application) or are they minimized (like the white color icons in example image) and if you wanna customize only one of the both states you can with `.minimized` and `.not-focused`. If you do not need different styling for visible and minimized windows, styling .running is sufficient.

PR #1077, opened 2026-08-24
feat(widget): add lock keys indicator

### LINKED ISSUE #1023: Digital indicator for caps lock, num lock
### What is the request?

yeah most keyboards have it and most bars dont have this feature, but it does look good for bars to support a icon like ▢/▣ or ▫/▪ etc to display states..

### Why is it useful?

for seeing info onscreen without looking down at keyboard

### How could it be implemented?

a module for this exclusively would be ok but i feel:
a widget module called text whoose label field support variables like `${numlock_state}` and a change in state trigger its update..

### Additional context

_No response_

iremsu-umit35:
## Summary

Adds a new Lock Keys widget for displaying the current Caps Lock and Num Lock states in the YASB bar.

Closes #1023

The widget:
- displays configurable Caps Lock and Num Lock state labels
- updates automatically when the keyboard lock state changes
- supports primary and alternate labels
- provides state-specific CSS classes for styling
- includes configuration documentation

## Testing

Tested locally on Windows with YASB.

- Verified Caps Lock ON/OFF state updates
- Verified Num Lock ON/OFF state updates
- Verified both states update correctly together
- Verified alternate label toggle
- `ruff check .` passes
- `ruff format --check .` passes
- pre-commit checks pass

## AI usage

AI tools were used as an assistive aid during implementation. The resulting changes were manually reviewed, tested locally, and validated with the project's linting and formatting tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ choco install yasb
 | [Language](https://github.com/amnweb/yasb/wiki/(Widget)-Language) | Shows the current input language and allows switching between languages. |
 | [Launchpad](https://github.com/amnweb/yasb/wiki/(Widget)-Launchpad) | A customizable launchpad for quick access to applications. |
 | [Libre Hardware Monitor](https://github.com/amnweb/yasb/wiki/(Widget)-Libre-HW-Monitor) | Connects to Libre Hardware Monitor to get sensor data. |
+| [Lock Keys](https://github.com/amnweb/yasb/wiki/(Widget)-Lock-Keys) | Displays the current Caps Lock and Num Lock states. |
 | [Media](https://github.com/amnweb/yasb/wiki/(Widget)-Media) | Displays media controls and information. |
 | [Media Lite](https://github.com/amnweb/yasb/wiki/(Widget)-Media-Lite) | A vertical and minimal album-style media widget. |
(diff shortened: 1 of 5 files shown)


amnweb:
You can use a custom widget for that. Why do you need a separate widget for a simple job?

iremsu-umit35:
Thanks for the feedback. I went with a separate widget because I wanted the lock states to be read directly from Windows and update automatically without requiring users to configure an external command. I also thought having configurable state labels and state-specific CSS classes would make it easier to use and style.

That said, if you prefer this to be implemented through the existing CustomWidget instead, I'm happy to rework it that way.

amnweb:
Main issue for me is the timer. 200ms is 5 polls a second, and widgets are built per bar, so on a 3 monitor setup that's 3 timers doing the same two GetKeyState calls forever. The 50ms lower bound makes it worse.

Maybe we can use our own hotkey listener for this instead of polling, we already have one in `core/utils/win32/hotkeys.py`. Worth looking into.

Also whatever it ends up being, it should be shared between bars instead of one instance per widget, see how CpuWidget keeps CpuWorker as a class attribute.

And when I say a custom widget can do the same job, I mean you can set up a custom widget to call a PowerShell command every second and give you the same result you have now with the new widget.

PR #1100, opened 2026-09-09
feat: add Windows Start Menu widget

darco-ctrl:
## Changes

- Adds a Windows Start Menu widget
- Uses the existing `start_menu()` Win32 utility
- Uses `BaseWidget`'s callback system
- Adds configurable icon support
- Tested locally
- Ruff/pre-commit checks pass


amnweb:
We really don't need a widget like this one. You can easily use this command in a custom widget.

Issue #1053, opened 2026-08-06
komorebi workspaces widget stops responding to clicks after display sleep/wake (stale cached monitor_hwnd)


dylanm5017:
### Windows version

Windows 11

### Windows OS build

26200.8875 (25H2)

### YASB version installed

YASB Reborn v2.0.5 x64 (stable), YASB-CLI v1.1.7

### Describe the bug

After the monitors sleep and wake, the `komorebi.workspaces` widget stops responding to clicks. The bar renders fine and the workspace buttons still highlight as the focused workspace changes, but clicking a button does nothing and each click logs `TypeError: 'NoneType' object is not subscriptable`.

komorebi itself is healthy throughout — `komorebic state` responds normally and its monitor list is correct. Only YASB needs restarting; `yasbc reload` fixes it until the next sleep.

**Steps to reproduce**

1. Run YASB with a `komorebi.workspaces` widget on a bar.
2. Let the displays sleep (or turn them off), then wake them.
3. Click any workspace button.

**Root cause**

`_komorebi_screen` is `None` because the widget is resolving its screen with a monitor handle that no longer exists.

- `src/core/bar.py#L105` caches the bar's monitor handle once in `__init__` and hands it to every widget at `#L262`:

  ```python
  self.monitor_hwnd = get_monitor_hwnd(int(self.winId()))   # L105, __init__ only
  ...
  widget.monitor_hwnd = self.monitor_hwnd                   # L262, _add_widgets only
  ```

- Windows reissues `HMONITOR` handles when the displays come back, so that cached value goes stale. `on_geometry_changed` (`bar.py#L152`) calls `position_bar()` and `update_app_bar()` but never recomputes `monitor_hwnd` — and it wouldn't help here anyway, since the resolution is unchanged and `geometryChanged` doesn't necessarily fire on a wake.

- `src/core/widgets/komorebi/workspaces.py#L403` then prefers that stale cached value over a live lookup:

  ```python
  self._screen_hwnd = self.monitor_hwnd or get_monitor_hwnd(int(QWidget.winId(self)))
  ```

  Because `self.monitor_hwnd` is always truthy, the live lookup is never reached, `get_screen_by_hwnd` (`#L406`) returns `None`, and the widget can never recover on its own. Every subsequent click hits `self.parent_widget._komorebi_screen["index"]` (`#L79` and `#L197`) and throws.

[...]

Issue #1050, opened 2026-08-05
LeopardWM integration


leifniem:
### What is the request?

https://github.com/jcardama/LeopardWM integration for workspaces (and maybe pause mode) as a widget.

### Why is it useful?

LeopardWM is a Windows counterpart to scrolling horizontal window managers like niri.
It supports features like workspaces and pause mode and would make for a nice addition to Yasb.

### How could it be implemented?

LeopardWM provides a `lwm subscribe` command, that issues data about events, as well as dedicated commands for all actions, that could be called from the widget.

amnweb:
Is that app stable enough for daily usage? And how does this scrolling work with multiple screens?

leifniem:
I am at least using it daily on my windows machine and found it to be more straight forward than komorebi or Glaze in terms of configuration and clearly focused on the scrolling layout.

I cannot speak to it's multi monitor usage, since I run a single display, but there are actions and event info for handling multiple monitors.

Mihir-Null:
Opened #1112 implementing the workspaces half of this — a native `leopardwm.workspaces` widget rather than custom widgets shelling out to `lwm`. Pause mode isn't included.

@amnweb, on your multi-monitor question: each bar resolves its own native `HMONITOR` and targets `lwm workspace N --monitor DEVICE` against that device, rather than matching on Qt's friendly display names. That matters because friendly names aren't unique — two identical monitors report the same model string, and name-based matching sends the switch to the wrong screen. There's also an explicit device override and an all-monitors mode.

I should be straight about the limits of my own testing there: the multi-monitor routing has synthetic coverage, but my live check ran on a single logical display. @leifniem is better placed to speak to daily-use stability, and I'd value them trying the branch on real hardware.

One dependency worth flagging before anyone invests review time: the widget needs a `workspace_state` subscription and a `--monitor` flag that LeopardWM doesn't ship yet. Those are proposed in jcardama/LeopardWM#118 and aren't settled, so #1112 shouldn't merge ahead of that. Against current LeopardWM releases the widget shows an offline state rather than misbehaving.

Issue #1033, opened 2026-07-24
systray widget's legacy monitor (use_hook: false) interferes with GlazeWM's window cloaking (Alt+Tab shows windows from all workspaces)


RiceNoodlee:
### Windows version

Windows 11

### Windows OS build

26200.8875

### YASB version installed

v2.0.5

### Describe the bug

This bug is really just a minor inconvenience 
When running yasb alongside GlazeWM (hide_method: 'cloak'), having the systray widget enabled with use_hook: false (the legacy systray monitor) causes Windows' native Alt+Tab switcher to show windows from all GlazeWM workspaces, not just the currently focused one — even though GlazeWM is correctly configured to hide/cloak windows on inactive workspaces.

Setting use_hook: true (the newer systray hook) resolves the issue entirely, with the systray widget otherwise working normally. Fully removing the systray widget also resolves it, confirming the widget (specifically its legacy mode) as the source.

**Steps to reproduce**

1. Run GlazeWM with:
yaml
   general:
     hide_method: 'cloak'
     show_all_in_taskbar: false
2. Run yasb with the systray widget enabled, using use_hook: false:
yaml
   systray:
     type: "yasb.systray.SystrayWidget"
     options:
       use_hook: false
       # (see attached full config)
3. Open windows across multiple GlazeWM workspaces, switch between workspaces a few times, then press Alt+Tab — windows from other workspaces are visible in the switcher.
4. Change only use_hook: false → use_hook: true in the systray widget options, reload yasb (GlazeWM left running the whole time, not restarted). Repeat step 3 — Alt+Tab now correctly shows only current-workspace windows.
5. As a further check, fully removing the systray widget (instead of toggling use_hook) also resolves it.

This was confirmed consistently across multiple repeats: use_hook: false → bug reproduces every time; use_hook: true or systray removed → bug does not occur.

**Expected behavior**

The systray widget (in either mode) should not affect GlazeWM's window cloaking/visibility state. Alt+Tab should only show windows from the current workspace regardless of the widget's configuration.

**Suspected cause**

The legacy systray monitor path (use_hook: false) appears to interact with the same OS-level state GlazeWM's cloak hiding method relies on (DWM cloak state / taskbar tab registration), effectively undoing or interfering with GlazeWM's per-workspace window hiding. The newer hook-based implementation (use_hook: true) does not have this conflict.

**Suggested fix**

[...]

Issue #738, opened 2026-02-24
glazewm: Empty workspaces are always hidden after 1.9.0


Cynosphere:
### Windows version

Windows 10

### Windows OS build

19044.6809

### YASB version installed

1.9.0

### Describe the bug

Empty workspaces are always hidden

Setting the new `monitor_exclusive` field to `false` shows no changes in behavior

### Relevant log output

```shell
N/A
```

PR #1112, opened 2026-09-14
feat(leopardwm): add native workspace widget using shared IPC snapshots

### LINKED ISSUE #784: YASB freezes when enabling app_icons in glazewm_workspaces
### Windows version

Windows 11

### Windows OS build

22631.6199

### YASB version installed

1.9.0 x64 (stable)

### Describe the bug

When I enable either the active icon or the populated icons, the bar refreshes, as soon as it detects GlazeWM it freezes. I checked the log file and didn't see anything related to GlazeWM in it.

Here's my relevant settings:

```yaml
bars:
  primary-bar:
   # ...
    widgets:
      left:
        - "home"
        - "glazewm_workspaces"
        - "glazewm_tiling"
        - "glazewm_binding"
        - "active_window"
      center:
        - "clock"
      right:
        - "media"
        - "weather"
        - "systray"
        - "notifications"
        - "power_menu"
widgets:
  # ...
    glazewm_workspaces:
    type: "glazewm.workspaces.GlazewmWorkspacesWidget"
    options:
      offline_label: "GlazeWM Offline"
      hide_empty_workspaces: true
      monitor_exclusive: true
      enable_scroll_switching: false
      app_icons:
        enabled_populated: false
        enabled_active: true
        size: 14
        max_icons: 0
        hide_label: false
        hide_duplicates: true
        hide_floating: false
# ...
```

### Relevant log output

```shell
???
```

Mihir-Null:
Implements #1050.

> **Stacked on #1125.** The shared `app_icons.py` fix that used to be part of this PR now lives in #1125 at review request. It is still the first commit in this branch so that this PR stays reviewable and testable on its own; once #1125 lands, a rebase drops it from this diff automatically. Review #1125 separately — it has no LeopardWM dependency.

## Summary

Adds `leopardwm.workspaces.WorkspaceWidget`: one native workspace strip per bar, backed by a shared event subscription rather than per-workspace custom widgets polling a persisted state file. Complete membership allows inactive and empty workspace filtering, accurate population, application icons, and switching the workspace on the clicked monitor.

In #1050 @amnweb asked how the scrolling layout behaves across multiple screens. The answer here is that the widget never infers a monitor from a friendly name: each bar resolves its own native `HMONITOR` and routes `lwm workspace N --monitor DEVICE` to that device, so duplicate monitor model names and Qt display names cannot cross-target. An explicit device override and an all-monitors mode are both available. @leifniem, as the requester and a daily LeopardWM user, is best placed to confirm the runtime behaviour on real hardware.

**Upstream dependency:** https://github.com/jcardama/LeopardWM/pull/118 supplies the proposed `workspace_state` subscription, one-shot query, and monitor-targeted command. This widget requires those capabilities; existing LeopardWM releases without them show an offline state. The consumer contract may need amendment following LeopardWM maintainer review and release/version reconciliation. **This should not merge before that contract settles.**

## Changes

- Shared, reference-counted Qt `QProcess` client per `lwm_path`, running `lwm subscribe --events workspace_state`. Multiple bars share one subscription; the last widget's disposal or YASB shutdown stops its children.
- Bounded UTF-8 NDJSON parsing and atomic begin/chunk/end snapshot assembly. Disconnects, malformed/partial snapshots, lag, and timeouts clear stale state and reconnect with backoff. Recognises the proposed standalone v3 and locally combined v4 snapshot formats; a version number alone is not treated as proof of the capability.
- Separate bounded command processes call `lwm workspace N --monitor DEVICE`, matched on native `HMONITOR`.
- Reuses YASB's native Windows icon/process lookup in background batches with a membership-scoped cache. Native, grayscale, and executable-to-font-glyph modes require no stored icon folder and no new runtime dependency.
- Adds the Pydantic model, widget documentation/sidebar entry, and explicit frozen-build package inclusion.

Module layout follows the existing conventions: `core/widgets/leopardwm/` beside `komorebi`/`glazewm`, `core/widgets/services/leopardwm/` beside the other service clients, and `core/validation/widgets/leopardwm/` beside the other validation models.

`schema.json` is intentionally **not** modified, per `docs/Contributing.md`. Running `python src/core/validation/export_schema.py` locally produces a clean additive diff (the LeopardWM models only, nothing removed or reordered), so the release-time `update-schema.yaml` regeneration will pick it up without conflict.

### Configuration options

| Area | Options |
| --- | --- |
| Connection/monitor | `lwm_path`, `monitor`, `monitor_exclusive` |
| Visibility | `show_inactive_workspaces`, `hide_empty_workspaces`, `hide_if_offline`, `label_offline` |
| Labels | Empty/active/populated label templates with `{index}`, `{name}`, `{count}`, `{monitor}` |
| Interaction | `enable_scroll_switching`, `reverse_scroll_direction` |
| `app_icons` | `enabled`, `size`, `max_icons`, `hide_label`, `hide_duplicates`, `hide_floating`, `monochrome`, `focused_monochrome`, `inactive_monochrome`, `cell_width`, `inactive_cell_width`, `mode`, `glyphs`, `fallback_icon` |
| Presentation | `workspace_separator`, `label_workspace_empty_btn`, `show_focus_indicator` |

Active workspaces remain visible when empty slots are hidden. Icon filters never change authoritative workspace population. Buttons expose active/focused/empty/populated CSS states and monitor/workspace tooltips. All new defaults preserve existing presentation.

## Related issues

These are prior reports in this repository that informed the design. Only #1050 is claimed to be resolved by this PR; the rest are listed as context and precedent, and several are already closed.

- **#1050 — LeopardWM integration** (open, `feature-request`). The originating request. This PR implements the workspace half; pause mode is not included.
- **#784 — YASB freezes when enabling `app_icons` in `glazewm_workspaces`** (closed) — now handled in **#1125**, not here. That PR bounds the unbounded `WM_GETICON` request in the shared `get_window_icon` helper and stops destroying the borrowed icon handle. It is not claimed to fix #784, which was never root-caused, but it removes a mechanism that produces that exact symptom and is still present on `main`. Discussion of that change belongs on #1125.
- **#1053 — komorebi workspaces stops responding to clicks after display sleep/wake (stale cached `monitor_hwnd`)** (closed). Direct precedent for the monitor-routing commit: a cached native handle that survives a display topology change becomes a silent no-op. Here the handle is re-resolved rather than cached across reconnects, and the offline path clears the cache. Covered by `test_offline_clears_buttons_and_current_handle_cache` and `test_native_handle_wins_over_screen_name_fallback`.
- **#540 — distinguish an active workspace from a focused workspace** (closed). Precedent for `show_focus_indicator`, `focused_monochrome` and `inactive_monochrome`: on multi-monitor setups every monitor has an active workspace, so "active" alone cannot indicate global focus. Those options default to the existing behaviour, so nothing changes unless opted into.
- **#738 — glazewm: empty workspaces are always hidden after 1.9.0** (closed). Precedent for the `hide_empty_workspaces` / `monitor_exclusive` semantics. The active workspace is deliberately retained when empty slots are hidden, so this widget cannot reproduce that class of "my workspace vanished" report.
- **#1033 — systray legacy monitor interferes with GlazeWM window cloaking** (open). Not addressed here; noted only because it also involves per-monitor window enumeration on Windows.

## Validation

Reviewers can reproduce the suite — it uses only the stdlib `unittest`, so no new dependency and no `pyproject.toml` change is required:

```pwsh
$env:PYTHONPATH = "$PWD\src;$PWD\tests"
python -m unittest discover -s tests -t tests -v
```

- **68 tests pass** on Windows / Python 3.14, covering real child-process lifecycle, invalid and partial streams, reconnects, monitor routing, clicks and scrolling, visibility, asynchronous icon results, glyph sizing, focus/cache transitions, separator lifecycle, accessible names, and shutdown cleanup.
- `ruff check .` and `ruff format --check .`: both clean across the tree.
- **Rebased on current `main` (2e87be51).** That surfaced one genuine failure worth noting: the focus-indicator test sampled a single pixel from the bottom row of the grabbed button and required the exact border colour. After `fix(dpi): handle display scale changes in bar and widgets`, `grab()` renders at the screen's device pixel ratio, so at 125% scaling a 2px logical border spans 2.5 physical rows and that last row is a half-covered blend (`#2f8e2f` over the `#1e1e1e` background). The underline itself was drawn correctly; the assertion encoded a 100%-scaling assumption. It now samples the whole `ceil(2 * ratio)` row band and asserts the colour is present when focused and absent when not, which is scale-independent and slightly stronger than before.
- Native x64 executable and MSI builds pass; the final archive contains the new widget, transport, state assembler, and validation module.
- Live personal deployment verified centered rendering, native grayscale icons, reload, clean stop/start, and exactly one subscription with no orphaned helper processes.

Known limits, stated plainly:

- Two-monitor behaviour has synthetic coverage only; the live check ran on one logical Windows display.
- ARM64 packaging has not been verified locally.
- CI on this PR has not run yet — the **PR Check** workflow is sitting at `action_required` because I am a first-time contributor here. It needs a maintainer to approve the workflow run before Ruff and the x64/ARM64 builds will report.
- This repository has no `tests/` directory today and no test runner in CI. I have added tests without wiring them into `pr-check.yaml`, since introducing a test stage is your call, not mine. If you want them run in CI I will add the job; if you would rather this contribution carry no tests at all, say so and I will drop them.

## AI assistance

Disclosed under the AI-Generated Code Policy in `docs/Contributing.md`. Substantial portions of the implementation, tests, and documentation were developed with AI assistance (OpenAI Codex), followed by local diff review and the automated and live verification recorded above. I have reviewed every line, I can explain any change in this diff on request, and the known limits above are stated so you can judge what has and has not actually been exercised.

## Review coordination

The linked LeopardWM PR asks @jcardama for guidance on opt-in subscription defaults, the `--events workspace_state` / `--monitor` flags, query/command shape, snapshot framing/publication, and protocol versioning alongside jcardama/LeopardWM#110. Feedback that changes that contract will be reflected here before merge.

Guidance welcome on the widget option names, lifecycle integration, and prerequisite/release policy.


Mihir-Null:
@amnweb — requesting review when you have time. I can't use the Reviewers field from a fork, so flagging it here instead.

Two things that need you specifically:

1. **The PR Check workflow is blocked at `action_required`** on all three runs so far, since this is my first PR here. Ruff and the x64/ARM64 builds won't report until you approve the run. Both Ruff checks are clean locally across the whole tree if that helps in the meantime.
2. **Please don't merge this yet.** It depends on jcardama/LeopardWM#118 for the `workspace_state` subscription and the `--monitor` flag, and that contract isn't settled. I'd rather land it after LeopardWM releases those, or restructure if the shape changes.

I've just updated the description to tie the work back to the existing issues, and removed an internal planning note that had no business being in `docs/` — that folder gets published verbatim to the wiki by `publish-wiki.yaml`, so it would have shipped as a public page.

This implements #1050. Your question there about multi-monitor behaviour is answered in the description: monitor routing is done on native `HMONITOR` rather than friendly names, so duplicate display model names can't cross-target.

@leifniem — you opened #1050 and mentioned you run LeopardWM daily. If you're willing to try this branch, you'd be covering the exact gap in my testing: I verified two-monitor routing synthetically but only had one logical display live. You're also better placed than I am to answer @amnweb's stability question from that thread.

@Video-Nomad — one piece may interest you given you triaged #784. That issue was closed when the reporter said it stopped reproducing after a reboot, so the cause was never found, but `get_window_icon` on `main` still calls `win32gui.SendMessage(hwnd, WM_GETICON, ...)` with no timeout (`app_icons.py:118`). That blocks the caller for as long as the target window's message loop is hung, which matches the reported freeze. The same function also calls `DestroyIcon` on the returned handle (`app_icons.py:125`), which is borrowed from the owning window and not ours to free.

I'm not claiming that fixes #784 — no root cause was ever established there. But both look wrong on `main` today and affect every caller of that shared helper, not just this widget. It's isolated in one commit (48176a4) with a regression test against a genuinely unresponsive HWND, and I'm happy to pull it into its own PR if you'd rather review it separately from the LeopardWM work. That one has no upstream dependency and could land independently.

On testing: the repo has no `tests/` directory today, so I've added one without touching `pyproject.toml` or `pr-check.yaml`. The tests are stdlib `unittest` only — no new dependency:

```pwsh
$env:PYTHONPATH = "$PWD\src;$PWD\tests"
python -m unittest discover -s tests -t tests -v
```

68 pass. Introducing a test stage is your call rather than mine, so I've left CI alone — glad to wire up a job, or drop the tests entirely, whichever you prefer.

AI assistance is disclosed in the description per `docs/Contributing.md`. Happy to walk through the reasoning behind any part of this diff.

Mihir-Null:
Updated per the review notes above.

**Split.** The shared `app_icons.py` change is now its own PR, #1125. @Video-Nomad — that's where the `WM_GETICON` discussion should go, rather than here. It has no LeopardWM dependency and can land on its own.

It remains the first commit on this branch so this PR stays independently reviewable and its tests stay green; once #1125 merges, a rebase drops it from this diff with no action needed. Its test moved to `tests/test_app_icons.py`, since the old name implied a LeopardWM connection it never had.

**Rebased** onto current `main` (2e87be51), and the two internal planning-doc commits are gone from the series entirely rather than being added and then removed. Five commits now: the shared icon fix, then the four LeopardWM ones.

The rebase caught a real failure I should flag rather than quietly fix. `test_focus_indicator_css_draws_underline_without_changing_geometry` read a single pixel from the bottom row of the grabbed button and required the exact border colour. After `fix(dpi): handle display scale changes in bar and widgets`, `grab()` renders at the screen's device pixel ratio — so at 125% scaling a 2px logical border spans 2.5 physical rows, and that last row is a half-covered blend of the border over the background (`#2f8e2f` against `#1e1e1e`). The underline was drawn correctly; my assertion had a 100%-scaling assumption baked into it. It now samples the whole `ceil(2 * ratio)` row band and checks the colour is present when focused and absent when not, which is scale-independent and a little stronger than the original.

Worth knowing generally: `QWidget.grab()` is DPI-aware after that commit, so any pixel-exact test against a grabbed widget needs to account for the ratio.

68 tests pass on the rebased branch; `ruff check .` and `ruff format --check .` are both clean.

@amnweb — the **PR Check** workflow is still at `action_required` on both this PR and #1125 and needs your approval before CI will report. #1125 is the smaller of the two and doesn't depend on anything upstream, so it may be the easier one to start with.

PR #1125, opened 2026-09-19
fix(icons): bound window icon requests and preserve borrowed handles

Mihir-Null:
Split out of #1112 at review request, since it changes a shared helper and has nothing to do with LeopardWM. It stands alone, has no upstream dependency, and can land independently of that PR.

## The problem

`get_window_icon` asks each window for its icon here:

```python
hicon = win32gui.SendMessage(hwnd, win32con.WM_GETICON, which, 0)
```

`SendMessage` does not return until the target window's thread services its message queue. A window that is hung, busy, or simply not pumping messages blocks the caller for as long as it takes — there is no upper bound. Callers that walk many windows, such as the `app_icons` paths in the workspace widgets, are exposed to the worst window in the set.

Separately, the result is passed to `DestroyIcon`:

```python
# WM_GETICON returns an icon handle we should destroy
win32gui.DestroyIcon(hicon)
```

That comment is not correct. `WM_GETICON` returns a handle that stays owned by the window it was read from — it is borrowed, not transferred. Destroying it can invalidate an icon the source window is still using. (`DestroyIcon` at line 410 is a different case and is left alone: that handle comes from `SHGetStockIconInfo`, which does transfer ownership.)

## The change

- `SendMessageTimeoutW` with `SMTO_ABORTIFHUNG | SMTO_BLOCK` and a 200ms bound per icon variant, so a hung window costs at most 600ms across the three variants instead of blocking indefinitely.
- The result is read through a `ctypes.c_size_t`. `WM_GETICON` returns a pointer-width value, so a `c_long` would truncate the `HICON` on 64-bit Windows.
- `DestroyIcon` is no longer called on the borrowed handle.
- The existing AUMID, window-class and default-icon fallbacks are untouched and still apply when a request times out, so a window that does not answer in time degrades to the same icon it would have got before.

## Relation to #784

#784 ("YASB freezes when enabling `app_icons` in `glazewm_workspaces`") was closed when the reporter found it no longer reproduced after a reboot, so no cause was ever identified. **I am not claiming this fixes it** — there is no way to confirm that against an issue that was never root-caused.

What I can say is that an unbounded `SendMessage` into arbitrary windows is a mechanism that produces exactly that symptom, it is reachable from the `app_icons` path named in that report, and it is still present on `main`. Whether or not it was the cause there, it is worth bounding.

## Tests

`tests/test_app_icons.py`, 7 tests, stdlib `unittest` only — no new dependency and no `pyproject.toml` change:

```pwsh
$env:PYTHONPATH = "$PWD\src;$PWD\tests"
python -m unittest discover -s tests -t tests -v
```

Coverage: pointer-width handles preserved on 64-bit, the borrowed handle not destroyed, each of the three timeout fallback paths (AUMID, class icon, default icon), the failed-send path leaving the output parameter unread, and a real unresponsive window — a private message-only `HWND` created on a separate thread that deliberately never pumps its queue — to confirm the wait is genuinely bounded rather than bounded only under mocks.

All 7 pass against this branch, and all 7 fail against `main` without the `app_icons.py` change, so they do hold the regression rather than merely passing alongside it.

Note that this repository has no `tests/` directory today, so this PR creates one. I have deliberately not touched `pr-check.yaml` — adding a test stage to CI is your call. Happy to wire up the job if you want it, or to drop the tests if you would rather not carry a test directory yet.

## Verification

- `ruff check .` and `ruff format --check .` both clean across the tree.
- Rebased on current `main` (2e87be51).

## AI assistance

Disclosed per the AI-Generated Code Policy in `docs/Contributing.md`. This change and its tests were developed with AI assistance (OpenAI Codex), then reviewed line by line and verified as recorded above. Happy to talk through any part of it.


Mihir-Null:
@amnweb @Video-Nomad — requesting review. I can't use the Reviewers field from a fork, so flagging it here.

Split out of #1112 because it changes a shared Win32 helper that every `app_icons` caller goes through, and that shouldn't be buried in a LeopardWM widget PR. It has no dependency on that work and can land on its own.

Short version: `get_window_icon` sends `WM_GETICON` with an unbounded `SendMessage`, so any window that isn't pumping its message queue blocks the caller indefinitely, and it then calls `DestroyIcon` on the returned handle, which is borrowed from the window rather than owned by us.

@Video-Nomad, you looked at #784 — that's the freeze-on-`app_icons` report that was closed when it stopped reproducing after a reboot. I'm not claiming this fixes it and I don't think that's knowable now, but an unbounded `SendMessage` into arbitrary windows produces exactly that symptom from exactly that code path, and it's still on `main`. You'd know better than I would whether that matches what you saw when you investigated.

The `PR Check` workflow is at `action_required` here as well and needs approval before CI reports. Ruff is clean locally across the tree, and the 7 tests pass on this branch and fail on `main` without the `app_icons.py` change, so they do hold the regression rather than just passing next to it.

Note this creates a `tests/` directory, which the repo doesn't have yet. I deliberately left `pr-check.yaml` alone — whether to add a test stage is your call. Happy to wire it up, or to drop the tests, whichever you prefer.

amnweb:
@Mihir-Null thanks for splitting this out. `win32gui` already has `SendMessageTimeout`, so this can stay simpler without the ctypes binding:

```py
for which in (win32con.ICON_BIG, win32con.ICON_SMALL, getattr(win32con, "ICON_SMALL2", 2)):
    try:
        _, hicon = win32gui.SendMessageTimeout(
            hwnd, win32con.WM_GETICON, which, 0, win32con.SMTO_ABORTIFHUNG | win32con.SMTO_BLOCK, 200
        )
    except Exception:
        break
```

`win32gui` raises on a timeout, so the `break` stops after the first one. If a window didn't answer `ICON_BIG` in time, it won't answer the other two either, so a busy window costs 200 ms instead of 600 ms.

One question though, how did you run into this in real use? Which apps were freezing? I can reproduce it by freezing a window on purpose, but I've never noticed it in normal use.

PR #1130, opened 2026-09-22
feat(volume): add VolumeProvider with interactive menu, device selection, and volume controls

BrainlessDip:
Implements VolumeProvider for Quick Launch service with:
- Configurable step size for volume adjustments
- Interactive menu with device selection
- Preview-only while typing, set volume on Enter
- Up/down volume icons and menu buttons
- Direct volume control via _set_volume_direct
- DPI-aware handling for display scale changes
- Adaptive rail behavior when no island is left


src/core/widgets/yasb/quick_launch.py
```diff
@@ -1287,6 +1287,10 @@ def _execute_result(self, index: int):
         if should_close is True:
             QTimer.singleShot(0, self._hide_popup)
         elif should_close is False and self._popup:
+            # Preserve scroll position so selection stays near the clicked item
```
BrainlessDip, line 1290: In this part of the code, if the provider's `execute` function returns `True`, the menu should close. If the `execute` function returns `False`, the menu should stay open.

Previously, when a provider's `execute` function returned `False`, selecting an item would refresh the results and move the cursor back to the top. This change preserves the current selection/scroll position instead, so the cursor stays where it was.

I've also attached a screenshot showing which providers return `False` from their `execute` function and when.

Image

